PR #1, opened 2022-06-14
Task v1

danylott:


Abnormaltype (approved):
looks good, all comments only about videos

PR #6, opened 2022-07-21
solution

DmytroSvirsa:


app/main.py
```diff
@@ -1,7 +1,40 @@
...
+        response = requests.get(url, params=parameters).json()
+    except TypeError:
+        print("you should give an api_key")
+        return
+
+
+
+    city = response["location"]["name"]
+    country = response["location"]["country"]
+    time = datetime.datetime.strptime(
```
SergeyDev1996, line 25: We could get the time from the API(by the key ['location']['localtime']), but if it works, than good

SergeyDev1996 (approved):

app/main.py
DmytroSvirsa: it seemed to me that it was somewhat illogical, to receive a forecast for a specific time, so I reform it as a date only

MasterpieceElbow (approved):

PR #25, opened 2022-07-24
Solution

RomaTsybulia:


app/main.py
```diff
@@ -1,6 +1,26 @@
...
+import requests
+
+
 def get_weather():
-    # write your code here
-    pass
+    base_url = "http://api.weatherapi.com/v1"
+    json = ".json?"
+    key = "API_KEY"
+    option = "/current"
```
TPolina, line 11: you can just save the full URL. No need to split it

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+    api_key = os.environ.get(key)
```
TPolina, line 12: it's okay to write os.environ.get("API_KEY"), variable key is unnecessary here

app/main.py
```diff
@@ -1,6 +1,26 @@
+import ast
+import os
+
+import requests
+
+
 def get_weather():
-    # write your code here
-    pass
+    base_url = "http://api.weatherapi.com/v1"
```
TPolina, line 8: constant variables better put out of function

TPolina (changes requested):

PR #34, opened 2022-09-07
'Solution'

LaskoA:
Please check my solution (image file is pretty big so please advice how to fix it). Thank you.


LaskoA:
Managed to decrease image size from 1GB to 400MB - please check it with next command:
"docker pull laskoa/docker-weather-api-slim"

Thank you.

LaskoA:
Now image is "only" 341Mb, please use next command:
"docker pull laskoa/docker-weather-api-slim-v3"

Thanks in advance.

.env
```diff
@@ -0,0 +1 @@
+API_KEY=7c508243647f4c4b971122001220609
```
DmytroSvirsa, line 1: that is your secret API_KEY, never push stuff like this,

DmytroSvirsa (changes requested):

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+RUN apt-get update \
+    && apt-get -y install libpq-dev gcc \
+    && pip install psycopg2
+
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "manage.py", "runserver"]
```
DmytroSvirsa, line 18: error python: can't open file '/app/manage.py': [Errno 2] No such file or directory

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+ENV PYTHONUNBUFFERED=1
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+
+RUN apt-get update \
+    && apt-get -y install libpq-dev gcc \
+    && pip install psycopg2
```
DmytroSvirsa, line 12: you can add it to requirements.txt

DmytroSvirsa (changes requested):

LaskoA:
Please check now after fixes, use command "docker pull laskoa/docker-weather-api" (old images were deleted from Docker).

P.S.: If you don`t mind I would leave next string in dockerfile (not in requirements.txt):
    RUN apt-get update \
    && apt-get -y install libpq-dev gcc \
    && pip install psycopg2

The thing is slim-image did not work without it so I would like to keep it seperately for the future as a reminder :)

Thank you.

DmytroSvirsa (approved):
Yep, well done!

PR #38, opened 2022-09-19
Solution

Splend1ed:


.dockerignore
```diff
@@ -0,0 +1,3 @@
+.gitignore
+README.md
+result.png
```
IvanRamyk, line 3: what about venv?

app/main.py
```diff
@@ -1,6 +1,18 @@
+import requests
+import os
+
+
 def get_weather():
-    # write your code here
-    pass
+    key = os.environ["API_KEY"]
```
IvanRamyk, line 6: could be a global constant

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    city = "Paris"
```
IvanRamyk, line 7: could be a global constant

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+    url = f"https://api.weatherapi.com/v1/current.json?key={key}&q={city}"
```
IvanRamyk, line 8: URL could be a global constant (or at least part without arguments)

requirements.txt
```diff
@@ -0,0 +1 @@
+requests==2.28.1
```
IvanRamyk, line 1: use `pip freeze > requirements.txt` to save versions for all dependencies

IvanRamyk (changes requested):
You haven't modify readme :(

requirements.txt
Splend1ed: But it works, if I add more requirements, the image will be bigger, right?

IvanRamyk (commented):
Good job!

IvanRamyk (approved):
Good job!

PR #39, opened 2022-09-21
Solution

DHushchyk:


.dockerignore
```diff
@@ -0,0 +1 @@
+venv
```
IvanRamyk, line 1: end of file

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ENV PYTHONUNBUFFERED 1
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
IvanRamyk, line 13: end of file

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:3.10.5
+LABEL maintainer="d.hushcyk@gmail.com"
```
IvanRamyk, line 2: The image of the service should be as thin as possible;

app/main.py
```diff
@@ -1,6 +1,20 @@
+import requests
+import os
+
+KEY = os.environ["API_KEY"]
+URL = f"https://api.weatherapi.com/v1/current.json?key={KEY}&q=Paris"
```
IvanRamyk, line 5: Create a constant to store a city
In this case, you can easily change it if you want

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #41, opened 2022-10-05
Solution

pavlejviki:


app/main.py
```diff
@@ -1,6 +1,27 @@
+import os
+
+import requests
+
+endpoint = "https://api.weatherapi.com/v1/current.json"
+api_key = os.environ.get("API_KEY")
+city = "Paris"
```
Abnormaltype, line 7: constants should be written in the CAPS LOCK)

requirements.txt
```diff
@@ -1,5 +1 @@
-flake8==5.0.4
-flake8-annotations==2.9.1
-flake8-quotes==3.3.1
-flake8-variables-names==0.0.5
-pep8-naming==0.13.2
+requests==2.27.1
```
Abnormaltype, line 1: don't remove these libraries

Abnormaltype (changes requested):

PR #46, opened 2022-10-31
Skantor

KantorSerhiy:


Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM python:3.10
```
IvanRamyk, line 1: > The image of the service should be as thin as possible;

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #49, opened 2022-11-05
Docker weather api

LavrykN:


DmytroSvirsa (changes requested):
You must modify the line with the correct command to pull your image in README

DmytroSvirsa (approved):

PR #58, opened 2022-11-25
Solution

StasOkhrym:


app/main.py
```diff
@@ -1,6 +1,41 @@
...
-    # write your code here
-    pass
+    response = requests.get(URL, {"key": API_KEY, "q": FILTERING})
+    data = response.json()
+
+    if response.status_code == 200:
+        city = data["location"]["name"]
+        region = data["location"]["region"]
+        localtime = datetime.strptime(
+            data["location"]["localtime"], "%Y-%m-%d %H:%M"
```
rostyak: Do we need "%Y-%m-%d %H:%M"? API gives us formatted data

app/main.py
```diff
@@ -1,6 +1,41 @@
...
+        )
+        temp_c = data["current"]["temp_c"]
+        last_updated = datetime.strptime(
+            data["current"]["last_updated"], "%Y-%m-%d %H:%M"
```
rostyak: Do we need "%Y-%m-%d %H:%M"? API gives us formatted data

rostyak (approved):

app/main.py
Y-Havryliv: so why did you approve? @rostyak

Y-Havryliv (changes requested):

app/main.py
Y-Havryliv: rostyak is right, you can leave just `data["location"]["localtime"]`

Y-Havryliv (changes requested):

Y-Havryliv (approved):

PR #60, opened 2022-11-26
Solution

wQuelS:


requirements.txt
```diff
@@ -3,3 +3,12 @@ flake8-annotations==2.9.1
...
 pep8-naming==0.13.2
+attrs==22.1.0
+certifi==2022.9.24
+charset-normalizer==2.1.1
+idna==3.4
+mccabe==0.7.0
+pycodestyle==2.9.1
+pyflakes==2.5.0
+requests==2.28.1
+urllib3==1.26.13
```
ketstap162, line 14: Wow, i think it's to many requirements... But it's not mistake... Maybe...

app/main.py
```diff
@@ -1,6 +1,27 @@
+import os
+
+import requests
+
+
+URL = "https://api.weatherapi.com/v1/current.json"
+API_KEY = os.environ.get("API_KEY")
+CITY = "Paris"
+PARAMS = {"key": API_KEY, "q": CITY}
```
ketstap162, line 9: `PARAMS` is not looks like a const.
I think you should use it like a variable in function `get_weather()`

requirements.txt
ketstap162: You can skip it

ketstap162 (approved):
Good job!

app/main.py
ketstap162: I think it's not a mistake, so i will approve it

TPolina (approved):

PR #63, opened 2022-11-26
Solution

kateryna3k:
don't judge harshly, I made a task with dotenv's help because in another way it doesn't work

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,30 @@
+from dotenv import load_dotenv
+import os
+import requests
+
+
+load_dotenv()
+
+URL = "https://api.weatherapi.com/v1/current.json"
+API_KEY = os.getenv("API_KEY")
+CITY = "London"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    response = requests.get(
+        URL,
+        params={"key": API_KEY, "q": CITY},
+    )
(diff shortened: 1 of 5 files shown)


Gr1zzZzzly (approved):

app/main.py
```diff
@@ -1,6 +1,30 @@
...
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    response = requests.get(
+        URL,
+        params={"key": API_KEY, "q": CITY},
+    )
+    print(response.json())
```
TPolina: delete unnecessary print

app/main.py
```diff
@@ -1,6 +1,30 @@
...
+    city = response.json()["location"]["name"]
+    country = response.json()["location"]["country"]
+    temp = response.json()["current"]["temp_c"]
+    date = response.json()["location"]["localtime"]
+    condition = response.json()["current"]["condition"]["text"]
```
TPolina: why do you repeat `json()` each time?

TPolina (changes requested):

app/main.py
kateryna3k: my bad( my program did't work and i tried different ways

TPolina (approved):
Well done!

PR #68, opened 2022-11-27
Solution

OvDima1:


rostyak (approved):
Nice, but what happens if we have an error from the weather API?

TPolina (approved):

PR #71, opened 2022-11-27
Created dockerfile for python weather api

alexzubkoff:


Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.10.4-alpine
+LABEL maintainer="alexzubkoff123@gmail.com"
+
+ENV PYTHONUNBUFFERED 1
+ENV API_KEY=8b89884a56104151964200013222711
```
mltdll, line 5: Pushing sensitive data such as this key to the source control is generally not the best idea. It might not be important here (and it might even be the intended solution for the task), but I'd suggest leaving your key secret and only creating the variable locally.

app/main.py
```diff
@@ -1,6 +1,15 @@
...
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    api_key = os.environ["API_KEY"]
+    print(f"Running with API key: {api_key}, weather in Paris now (JSON):")
+    weather_in_paris_current = requests.get(
+        "http://api.weatherapi.com/v1/current.json?"
+        f"key={api_key}&q=Paris"
+    )
```
mltdll, line 11: Did you check the checklist? :)
https://github.com/mate-academy/py-docker-weather-api/blob/master/checklist.md#1-use-constants

mltdll (changes requested):

PR #77, opened 2022-11-28
Solution

pazyn404:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
-    # write your code here
-    pass
+    params = {
+        "key": os.environ.get("API_KEY"),
+        "q": CITY,
+    }
+    data = requests.get(url=URL, params=params).json()
+
+    for key, value in data.items():
+        print(f"{key}: {value}")
```
wQuelS, line 17: I guess you had to implement similar output as it was in the task Readme file

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
wQuelS, line 14: newline would be nice here :)

wQuelS (changes requested):

PR #79, opened 2022-11-28
Solution

KatrinLazarenko:


app/main.py
```diff
@@ -1,6 +1,27 @@
+import os
+import requests
+
+
+URL = "http://api.weatherapi.com/v1/current.json/?"
+API_KEY = os.environ.get("API_KEY")
+FILTERING = "Paris"
+result = requests.get(URL + f"q={FILTERING}")
```
FranchukS: You need to add API_KEY to request parameters as key=, so that the request works correctly

FranchukS (changes requested):

app/main.py
KatrinLazarenko: With parameters was my first version but after checklist I decided to rewrite it...

FranchukS (approved):
Nice!

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+        location = f"{info['location']['name']}/{info['location']['country']}"
+        localtime = info["location"]["localtime"]
+        temperature = info["current"]["temp_c"]
+        condition = info["current"]["condition"]["text"]
+        print("Performing request to Weather API for city Paris...")
+        print(
+            f"{location} {localtime} "
+            f"Weather: {temperature} Celsius, {condition}"
+        )
+    except Exception:
```
TPolina: it's not really good to use Exception. it's better to specify it

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+URL = "http://api.weatherapi.com/v1/current.json/?"
+API_KEY = os.environ.get("API_KEY")
+CITY = "Paris"
+parameters = {"key": API_KEY, "q": CITY}
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    try:
```
TPolina, line 12: if you want to use try block, remember that you should write only code that can cause errors inside it.

TPolina (changes requested):

app/main.py
KatrinLazarenko: Here I decided to use try/except (for sure) because I use dict[key] not dict.get(key).

TPolina (approved):

PR #80, opened 2022-11-28
Solution

linaSam:


wQuelS (approved):

.dockerignore
```diff
@@ -0,0 +1,3 @@
+venv
+result.png
+checklist.md
```
Nattalli, line 3: add blank line

app/main.py
```diff
@@ -1,6 +1,25 @@
+import os
+import requests
+
+
+WEATHER_URL = "http://api.weatherapi.com/v1/current.json?"
+API_KEY = os.environ.get("API_KEY")
+city = "Paris"
+res = requests.get(WEATHER_URL + f"q={city}")
```
Nattalli, line 8: it's better to use `params` instead of f-string

Nattalli (changes requested):

PR #88, opened 2022-11-28
Solution

Vasyl-Poremchuk:


app/main.py
```diff
@@ -1,6 +1,25 @@
+import os
+
+import requests
+
+# API_KEY = "b0480265525c4b09844183826222811"
```
wQuelS: i guess this has to be removed.

app/main.py
```diff
@@ -1,6 +1,25 @@
...
-    # write your code here
-    pass
+    response = requests.get(URL).json()
+    print(
+        f"Country: {response['location']['country']}.\n"
+        f"City: {response['location']['name']}.\n"
+        f"Time zone: {response['location']['tz_id']}.\n"
+        f"Current time: {response['location']['localtime']}.\n"
+        f"Temperature: {response['current']['temp_c']} °C.\n"
+        f"Wind speed: {response['current']['wind_mph']} m/h."
```
wQuelS, line 20: I don't know if it is required to make output similar to that which is in Readme file, so it is up to you. You can skip this comment and I'll approve it anyway

wQuelS (changes requested):

PR #89, opened 2022-11-29
Solution

taras-andruschenko:


app/main.py
```diff
@@ -1,6 +1,18 @@
...
+
+URL = "https://api.weatherapi.com/v1/current.json"
+CITY = "Paris"
+KEY = os.environ.get("API_KEY")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    data = requests.get(URL + "?" + f"key={KEY}" + "&" + f"q={CITY}").json()
```
IvanStored, line 10: u can use `params` here, if u want)

IvanStored (approved):

app/main.py
TPolina: agree, use params or make this expression a separate constant

TPolina (changes requested):

PR #91, opened 2022-11-29
Solution

AriohBelskij:


FranchukS (approved):
👍

.dockerignore
```diff
@@ -0,0 +1 @@
+venv/
```
Y-Havryliv: add blank line

Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:3.10
```
Y-Havryliv: `The image of the service should be as thin as possible;`

Y-Havryliv (changes requested):

Nattalli (approved):
👍🏻

PR #95, opened 2022-12-14
Solution

InnaKushnir:


app/main.py
```diff
@@ -1,6 +1,18 @@
+import os
+
+import requests
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+
+    API_KEY = os.environ.get("API_KEY") # noqa
```
IvanRamyk: why use `noqa` here?

IvanRamyk: Let's move constants to the top of the file

IvanRamyk (changes requested):

andrewbelo (approved):

PR #125, opened 2023-03-15
Weather API

Paul-Starodub:


app/main.py
```diff
@@ -1,6 +1,28 @@
+import requests
+import os
+from dotenv import load_dotenv
+
+load_dotenv()
+
+URL = "https://api.weatherapi.com/v1/current.json"
+QUERY = {"key": os.getenv("API_KEY"), "q": "Paris"}
```
Bohdan-Salamakha, line 8: place constant(s) and load_dotenv() into class/function or program start-point to not create redundant imports at the unexpected program places

Bohdan-Salamakha (changes requested):
v1

app/main.py
Paul-Starodub: Variable in function should be lowercase

Bohdan-Salamakha (approved):

PR #140, opened 2023-04-08
Solution

MaybeTI:
image


SergiiMachulin (approved):

app/main.py
```diff
@@ -1,6 +1,26 @@
+import os
+import requests
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    API_KEY = os.environ.get("API_KEY")
+    CITY = "Paris"
+    URL = "http://api.weatherapi.com/v1/current.json?"
```
YuriiKindrat: it is better to create constants outside of funcs

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+
+    response = requests.get(f"{URL}key={API_KEY}&q={CITY}")
```
YuriiKindrat: it is better to avoid using f-string here, u can use params arg instead

YuriiKindrat (changes requested):
image
fix this

wQuelS (approved):

PR #146, opened 2023-04-11
Solution

SergiiMachulin:


kuznetsb (changes requested):
Read last paragraph in README, you should change there path to your docker image so that I and mentors can test it.

wQuelS (approved):
Good

PR #148, opened 2023-04-12
'Solution'

Morty67:


Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:3.11.0-slim
+LABEL maintainer="gkarabetskii@gmail.com"
+
+ENV PYTHONUNBUFFERED 1
+
+WORKDIR /app
+
+COPY requirements.txt .
+RUN pip install -r requirements.txt
```
DmytryTong, line 9: I would like to change COPY string to COPY requirements.txt requirements.txt, but if it's pushing well - may be ok)

DmytryTong (approved):
I've made my notice)

wQuelS (approved):
GJ

PR #152, opened 2023-04-12
solution

amber-marichi:
docker pull maxmarichi/docker-weather-api


Maytreyaya (approved):

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+
+URL = "http://api.weatherapi.com/v1/current.json"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+
+    key = os.environ.get("API_KEY")
+    city = "Paris"
```
wQuelS: maybe these two variables can be as constants

wQuelS (changes requested):

app/main.py
amber-marichi: got it, thank you

wQuelS (approved):
Nice

PR #153, opened 2023-04-12
Solution

bihunva:


yura-hudzovskyi (approved):
GJ!

app/main.py
```diff
@@ -1,6 +1,41 @@
+import os
+
+import requests
+
+API_KEY = os.environ.get("API_KEY")
+if not API_KEY:
+    raise ValueError("API_KEY not found in environment variables")
```
wQuelS: you don't need to handle this

app/main.py
```diff
@@ -1,6 +1,41 @@
...
+        print(
+            f"{country}/{CITY} {_datetime} "
+            f"Weather: {temperature} Celsius, {condition}"
+        )
+
+    except requests.exceptions.HTTPError as err:
+        print(f"HTTP error occurred: {err}")
+
+    except requests.exceptions.RequestException as err:
+        print(f"An error occurred: {err}")
```
wQuelS: there is also no need to handle those errors in our case

wQuelS (changes requested):

wQuelS (approved):
Good, as always)

PR #156, opened 2023-04-13
Solution

yura-hudzovskyi:


NazarHladaniuk (approved):

app/main.py
```diff
@@ -1,6 +1,32 @@
+import os
+import requests
+
+API_KEY = os.environ.get("API_KEY")
+if not API_KEY:
+    raise ValueError("API_KEY not found")
```
wQuelS: yo don't have to handle this here

app/main.py
```diff
@@ -1,6 +1,32 @@
...
+FILTERING = "Paris"
```
wQuelS: filtering is not that good name here

wQuelS (changes requested):

wQuelS (approved):
GJ

PR #157, opened 2023-04-13
solution

DavidGlusch:


VladyslavCherkashyn (approved):

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+import os
+
+WEATHER_API_URL = "http://api.weatherapi.com/v1/current.json"
+API_KEY = os.getenv("API_KEY")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    payload = {"key": API_KEY, "q": "SW1"}
```
wQuelS: what is "SW1" ?:)

app/main.py
```diff
@@ -1,6 +1,16 @@
...
+
+    response = requests.get(WEATHER_API_URL, payload)
+
+    print(response.content)
```
wQuelS: can you make it more user-friendly?

wQuelS (changes requested):

app/main.py
DavidGlusch: nmmhjmhjm
i wanted to be mysterious :D

.env-example
```diff
@@ -0,0 +1 @@
+API_KEY=your weatherapi.com api key
```
Serhii-Leonenko: better to use .env.sample
also add a blank line

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):

PR #159, opened 2023-04-13
Solution

ada-krav:


requirements.txt
```diff
@@ -3,3 +3,4 @@ flake8-annotations==2.9.1
 flake8-quotes==3.3.1
 flake8-variables-names==0.0.5
 pep8-naming==0.13.2
+requests~=2.28.2
```
NEkropo1, line 6: does it work? I mean I see this ~= first time )

NEkropo1 (approved):
LGTM!

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+
+URL = "https://api.weatherapi.com/v1/current.json"
+FILTERING = "Paris"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    print("Performing request to Weather API for city Paris...")
+    api_key = os.getenv("API_KEY")
```
wQuelS: make it constant, please:)

app/main.py
```diff
@@ -1,6 +1,22 @@
+import requests
+import os
+
+
+URL = "https://api.weatherapi.com/v1/current.json"
+FILTERING = "Paris"
```
wQuelS: it is just a `CITY` or `LOCATION` isn't it?:)

wQuelS (changes requested):

app/main.py
ada-krav: Well, it`s according to best practice from the checklist, so ... 
image

app/main.py
ada-krav: while I don't think that my weather key is something valuable (and I'm currently at the point where I hate env variables), it was in the requirements:
image

app/main.py
wQuelS: I guess we need to update checklist:)

app/main.py
wQuelS: It is about following best practices. The same as u've done with `url` and `q`-param

wQuelS (changes requested):

app/main.py
NEkropo1: Actually I disagree, I mean when you filter queryset by exact parameter it's better to see `filter`, not `city` or `location`. Sure, best would be something like filter_by_city, but it's pretty redundant since we apply only 1 filter)

YuriiKindrat (approved):

PR #160, opened 2023-04-13
Solution

NEkropo1:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    params = {"key": API_KEY, "q": FILTER}
+
+    response = requests.get(URL, params=params)
+
+    print(response.content)
```
Yulia-Mala: emmm .. it's raw json,)) I guess the task implies that we should make some formatting)

Yulia-Mala (changes requested):

app/main.py
NEkropo1: Idk, I think its ok to return raw json with full data, depends on what we gonna get from it.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
wQuelS: +new line:)

app/main.py
```diff
@@ -1,6 +1,21 @@
+import os
+from pprint import pprint
+
+import requests
+
+
+API_KEY = os.getenv("API_KEY")
+
+FILTER = "Paris"
```
wQuelS: filter is not that good name here

wQuelS (changes requested):

Dockerfile
NEkropo1: sure, sry )

wQuelS (approved):

PR #162, opened 2023-04-13
Solution

alexlega:


app/main.py
```diff
@@ -1,6 +1,17 @@
+import requests
+
+API_KEY = "c4c72a4109d14453bd0133710231304"
```
WDemidenko: According to Readme: "You need to generate API_KEY for using Weather API, but it shouldn't be hard-coded in main.py. You should use environment variables for that purpose. So, read API_KEY from environment. And to pass environment variables to docker container use -e flag."

WDemidenko (changes requested):

wQuelS (approved):
Good. But yeah, never leave such sensitive info like API_keys at your repo

PR #163, opened 2023-04-13
Solution

NazarHladaniuk:


inrock1 (approved):

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
wQuelS: + new line:)

app/main.py
```diff
@@ -1,6 +1,26 @@
+import os
+
+import requests
+from typing import Dict
+
+API_KEY = os.environ["API_KEY"]
+FILTERING = "Paris"
```
wQuelS, line 7: filtering is not that good name here

wQuelS (changes requested):

app/main.py
NazarHladaniuk: image

NazarHladaniuk: In the beginning, I want to use CITY but...

wQuelS: Guess we have to update `checklist` :))

wQuelS (approved):
GJ

PR #164, opened 2023-04-13
Solution

VladyslavCherkashyn:


skmiwa1 (approved):

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM python:3.10.5
```
wQuelS: image

app/main.py
```diff
@@ -1,6 +1,27 @@
+import requests
+
+CITY = "Paris"
+URL = "http://api.weatherapi.com/v1/current.json"
+API_KEY = "3e243be0a1204288bfc132613231304"
```
wQuelS: what? Never ever put such sensitive data in such an insecure manner as you've done here. Use `.env`.
This is a very huge mistake

wQuelS (changes requested):

wQuelS (approved):
Ok, but consider using `python:3.11-alpine` or something similar

PR #172, opened 2023-04-13
Solution

inrock1:


app/main.py
```diff
@@ -1,6 +1,24 @@
...
+
+
+CITY = "Kiev"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    key = os.getenv("API_KEY")
+    url = "https://api.weatherapi.com/v1/current.json"
```
YuriiKindrat: this could be a constants too

app/main.py
```diff
@@ -1,6 +1,24 @@
+import os
+
+import requests
+
+
+CITY = "Kiev"
```
YuriiKindrat: image
huh ?

YuriiKindrat (changes requested):

app/main.py
inrock1: fixed

Serhii-Leonenko (approved):

PR #173, opened 2023-04-13
Solution with own functionality : u can use env.variable CITY for cho…

oigod:
…ose from what city u want see weather


.dockerignore
```diff
@@ -0,0 +1 @@
+venv
```
Serhii-Leonenko: add a blank line

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:3.11.1
```
Serhii-Leonenko: read the checklist

app/main.py
```diff
@@ -1,6 +1,39 @@
...
+
+import requests
+
+import exceptions
+from dotenv import load_dotenv
+
+load_dotenv("./.env")
+KEY = os.getenv("API_KEY")
+BASE_URL = "http://api.weatherapi.com/v1"
+WEATHER_ENDPOINT = "/current.json"
```
Serhii-Leonenko: This const is redundant add it to the URL

app/main.py
```diff
@@ -1,6 +1,39 @@
...
+CITY = os.getenv("CITY", "Paris")
+CITY_WEATHER_ENDPOINT = f"{BASE_URL}{WEATHER_ENDPOINT}?key={KEY}&q={CITY}"
```
Serhii-Leonenko: add query params with `params=` key in requests.get() method

Serhii-Leonenko (changes requested):
image

app/main.py
oigod: I write this by the zen of python:
 Explicit is better than implicit.

app/main.py
oigod: Can u pls be more accurate - i don`t understand u.

YuriiKindrat: it is better to avoid using f-strings with urls
u can add params argument to ur method get in 19th line, with dicts of params to ur url

app/main.py
YuriiKindrat: what about this point ? 
`Flat is better than nested.`
in context of this task, it is better to create only 3 base constant CITY, KEY, URL )

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):

PR #176, opened 2023-04-13
'Solution'

ghost:


app/main.py
```diff
@@ -1,6 +1,25 @@
...
+
+API_KEY = os.getenv("API_KEY")
+URL = "https://api.weatherapi.com/v1/current.json?"
+CITY = "Paris"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    result = requests.get(URL + f"key={API_KEY}&q={CITY}")
```
YuriiKindrat: no need to use f-string here, use params arg instead

YuriiKindrat (changes requested):
image
fix this

wQuelS (approved):

PR #179, opened 2023-04-14
'Solution'

oksanaaam:


app/main.py
```diff
@@ -1,6 +1,29 @@
...
+URL = "http://api.weatherapi.com/v1/current.json"
+FILTERING = "Paris"
+API_KEY = os.environ.get("API_KEY")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    response = requests.get(
+        f"{URL}?key={API_KEY}&q={FILTERING}&aqi=no")
```
YuriiKindrat: fix indentation here, also there is no need to use f-string here, u can use params arg instead

YuriiKindrat (changes requested):
image
fix this

app/main.py
```diff
@@ -5,19 +5,23 @@
...
 URL = "http://api.weatherapi.com/v1/current.json"
 FILTERING = "Paris"
-API_KEY = os.environ.get("API_KEY")
 
 
 def get_weather() -> None:
-    response = requests.get(
-        f"{URL}?key={API_KEY}&q={FILTERING}&aqi=no")
+    params = {
+        "key": os.environ.get("API_KEY"),
```
wQuelS: just make `API_KEY` as constant as it was:)

app/main.py
```diff
@@ -5,19 +5,23 @@
...
+        "q": FILTERING,
+        "aqi": "no"
```
wQuelS: why do you need this?

wQuelS (changes requested):

YuriiKindrat (approved):

PR #184, opened 2023-04-15
Solution

TeliksPG:


app/main.py
```diff
@@ -1,6 +1,23 @@
...
+
+API_KEY = os.environ.get("API_KEY")
+LOCATION = "Paris"
+URL = "http://api.weatherapi.com/v1/current.json?"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    response = requests.get(f"{URL}key={API_KEY}&q={LOCATION}")
```
YuriiKindrat: no need to use f-string here, it is better to use params arg instead

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):

PR #185, opened 2023-04-17
Solution

natalie-goriela:


Bohdan-Salamakha (changes requested):
I will not have to use your program without an API key
update your dockerhub image
image

Bohdan-Salamakha (approved):

PR #186, opened 2023-04-17
Solution

Anatolii-Poznyak:


app/main.py
```diff
@@ -1,6 +1,15 @@
...
+
+URL = "https://api.weatherapi.com/v1/current.json?"
+FILTERING = "Paris"
+API_KEY = os.environ.get("API_KEY")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    res = requests.get(URL + f"key={API_KEY}&q={FILTERING}")
```
YuriiKindrat: no need to use f-string here, it is better to use params arg instead

YuriiKindrat (changes requested):

wQuelS (approved):

PR #193, opened 2023-04-24
v1

MaxiTDI:


.dockerignore
```diff
@@ -0,0 +1,7 @@
+venv
+.env
+.flake8
+checklist.md
+result.png
+.pytest_cache/
+**__pycache__/
```
vsmutok: add new empty line

app/main.py
```diff
@@ -1,6 +1,29 @@
...
 def get_weather() -> None:
-    # write your code here
-    pass
+    response = requests.get(
+        url=URL,
+        params={
+            "key": API_KEY,
+            "q": CITY
+        }
+    ).json()
```
vsmutok: what happens if you get an error in response?


I think you should add exception handling

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #237, opened 2023-07-26
'Solution'

HalynaPetrova:


app/main.py
```diff
@@ -1,6 +1,26 @@
...
+    result = requests.get(BASE_URL, params={
+        "key": API_KEY,
+        "q": FILTERING})
+
+    if result.status_code == 200:
+        country = result.json()["location"]["country"]
+        city = result.json()["location"]["name"]
+        local_time = result.json()["location"]["localtime"]
+        temperature = result.json()["current"]["temp_c"]
+        condition = result.json()["current"]["condition"]["text"]
```
vsmutok: DRY 

now you do `result.json()` 5 times
Create separate variable for this

vsmutok: also use `get()`

requirements.txt
```diff
@@ -1,5 +1,18 @@
+asgiref==3.7.2
+certifi==2023.7.22
+charset-normalizer==3.2.0
+Django==4.2.3
```
vsmutok: why you need Django at requirements?

.dockerignore
```diff
@@ -0,0 +1,2 @@
+venv
+.python-version
```
vsmutok: add new empty line

vsmutok (changes requested):

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+
+    city = response.get("location")["name"]
+    country = response.get("location")["country"]
+    date = response.get("location")["localtime"]
+    temperature = response.get("current")["temp_c"]
+    condition = response.get("current")["condition"]["text"]
+
+    print(
+        f"{city}/{country} {date} "
+        f"Weather: {temperature} Celsius, {condition}")
```
vsmutok:
```suggestion
    response = requests.get(
        "http://api.weatherapi.com/v1/current.json",
        params={
            "key": os.environ.get("API_KEY"),
            "q": "Paris"
        }
    )
    data = response.json()

    city = data.get("location")["name"]
    country = data.get("location")["country"]
    date = data.get("location")["localtime"]
    temperature = data.get("current")["temp_c"]
    condition = data.get("current")["condition"]["text"]

    print(
        f"{city}/{country} {date} "
        f"Weather: {temperature} Celsius, {condition}"
    )
```

requirements.txt
```diff
@@ -1,5 +1,16 @@
...
+idna==3.4
+mccabe==0.7.0
 pep8-naming==0.13.2
+pycodestyle==2.9.1
+pyflakes==2.5.0
+pytz==2023.3
+requests==2.31.0
+sqlparse==0.4.4
+tzdata==2023.3
+urllib3==2.0.4
```
vsmutok: also, why you need all this requirements?
if you use only `requests`?

vsmutok (changes requested):

app/main.py
HalynaPetrova: Thank you))

requirements.txt
HalynaPetrova: I myself don’t understand where they all came from(( I accidentally added Django, because I opened the wrong project. How the others were added, I don’t understand((

HalynaPetrova: They are added automatically when I do: pip freeze > requirements.txt

vsmutok (approved):
good job 👍

PR #239, opened 2023-08-01
done

nazarbychkar:


app/main.py
```diff
@@ -1,6 +1,30 @@
...
-    # write your code here
-    pass
+    result = requests.get(
+        URL
+        + "?"
+        + f"key={API_KEY}"
+        + "&"
+        + f"q={FILTERING}"
+        + "&"
+        + "aqi=no"
```
ArsenPidhoretskyi: Looks so bad, 
FYI .get you can pass params as dict
Or write inline
```suggestion
        + f"{URL}?key={API_KEY}&q={FILTERING}&aqi=no"
```

ArsenPidhoretskyi (changes requested):

app/main.py
nazarbychkar: yeah, thx, mine version is something much over the top :)

vasylhnatiuk (approved):

PR #241, opened 2023-08-02
Solution

diana-shyrokikh:


app/main.py
```diff
@@ -1,6 +1,29 @@
...
+API_KEY = os.environ.get("API_KEY")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    result = requests.get(
+        URL
+        + f"?q={FILTERING}&key={API_KEY}"
+    )
```
YuriiKindrat: u can use `params` as param for get instead of f-string 
try to avoid using f-stings in such cases if it possible

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,6 +1,32 @@
...
+FILTERING = "Paris"
+API_KEY = os.environ.get("API_KEY")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    result = requests.get(
+        URL,
+        params={
```
LiudmylaKulinchenko: How about to make a global variable for these params?

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):

PR #246, opened 2023-08-05
Solution

imelnyk007:


Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ENV PYTHONUNBUFFERED 1
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
YuriiKindrat: no blank line at the end of file

app/main.py
```diff
@@ -1,6 +1,26 @@
+import os
+import requests
+from dotenv import load_dotenv
+
+load_dotenv()
+
+API_KEY = os.getenv("API_KEY")
+CITY = "Paris"
+URL = f"http://api.weatherapi.com/v1/current.json?key={API_KEY}&q={CITY}"
```
YuriiKindrat: no need to use f-string here
u can use `params` param to add them to the url

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,6 +1,30 @@
+import os
+import requests
+from dotenv import load_dotenv
+
+load_dotenv()
+
+API_KEY = os.getenv("API_KEY")
+CITY = "Paris"
+URL = f"http://api.weatherapi.com/v1/current.json"
```
Oleg-s-s: Check is failing here

Oleg-s-s (changes requested):

Oleg-s-s (approved):

PR #248, opened 2023-08-07
Soultion

YKonoplyov:


.dockerignore
```diff
@@ -0,0 +1,2 @@
+venv
+.env
```
Oleg-s-s: blank line

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:3.11.4-alpine3.17
+LABEL maintainer="kannabis252@gmail.com"
+
+ENV PYTHONUNBUFFERED 1
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
```
Oleg-s-s: Is there actual need to call `COPY` twice?

Oleg-s-s (changes requested):

Oleg-s-s (approved):

PR #255, opened 2023-08-07
solution

YevheniiIovchu:


.dockerignore
```diff
@@ -0,0 +1 @@
+venv
```
Oleg-s-s, line 1: Its seems to me like there are more files and folders you should ignore

Oleg-s-s (changes requested):

YuriiKindrat (approved):

PR #257, opened 2023-08-08
Solution

Gvilion:


.dockerignore
```diff
@@ -0,0 +1 @@
+venv
```
Oleg-s-s, line 1: it seems like there are more unnecessary files/folders

Oleg-s-s (changes requested):
image

.dockerignore
```diff
@@ -0,0 +1,3 @@
...
+.idea
+.env
```
LiudmylaKulinchenko: missed blank line

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 🐣

PR #258, opened 2023-08-08
done

maccarets:


app/main.py
```diff
@@ -1,6 +1,29 @@
+import requests
+import os
+
+from dotenv import load_dotenv
+
+load_dotenv()
```
YuriiKindrat, line 6: why do u need this ?

YuriiKindrat (changes requested):
what about this point ?
image

app/main.py
maccarets: to load environment variable from .env file

app/main.py
```diff
@@ -1,6 +1,29 @@
...
+
+URL = " https://api.weatherapi.com/v1/current.json"
+
+PARAMS = {
+    "key": os.getenv("API_KEY"),
+    "q": "Paris"
```
LiudmylaKulinchenko, line 12: How about to make those also global variables?

LiudmylaKulinchenko (approved):
Well done 🔥🐣

PR #259, opened 2023-08-08
'Solution'

yakhlamova:


app/main.py
```diff
@@ -1,6 +1,29 @@
+import requests
+import os
+from dotenv import load_dotenv
+
+load_dotenv()
```
YuriiKindrat: why do u need this ?

YuriiKindrat (changes requested):
what about this point ?)
image

LiudmylaKulinchenko (approved):
Good job! 🍬

PR #260, opened 2023-08-08
Solution

erikagayan:


app/main.py
```diff
@@ -1,6 +1,28 @@
+import os
+import requests
+
+
+CITY = "Paris"
+KEY = os.environ["API_KEY"]
+URL = f"https://api.weatherapi.com/v1/current.json?q={CITY}"
```
YuriiKindrat: it is better to use `params` param instead of f-string for urls params

YuriiKindrat (changes requested):
what about this point ?
image

LiudmylaKulinchenko (changes requested):
what about this point? 
`You must modify this line with correct command to pull your image: COMMAND=docker pull <YOUR_DOCKER_ID/YOUR_IMAGE_NAME>.`

LiudmylaKulinchenko (approved):
Good job! 😎🍬

PR #261, opened 2023-08-08
solution

OlhaTryhub:


Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ENV PYTHONBUFFERRED 1
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
YuriiKindrat: no blank line here (

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,7 +1,27 @@
+import requests
+
+
+API_KEY = "967cf4e71dfe436ba53110325230808"
```
YuriiKindrat: also did u read a task ? a user that pulled an image should add his own api_key through env vars )

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #262, opened 2023-08-08
'Solution'

RedCatHot:


Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ENV PYTHONUNBUFFERED=1
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
YuriiKindrat: no blank line here (

app/main.py
```diff
@@ -1,7 +1,37 @@
...
+        f"Temperature: {temp_c}°C, feels like {feelslike_c}°C.\n"
+        f"Wind: {gust_kph} km/h \n"
+        f"Humidity: {humidity}% \n"
+        f"Precipitation: {pressure_in} mm \n"
+        f"Last updated: {last_updated}"
+    )
 
 
 if __name__ == "__main__":
+    load_dotenv()
```
YuriiKindrat: why do u need this ?

app/main.py
```diff
@@ -1,7 +1,37 @@
...
+
+    API_KEY = os.environ.get("API_KEY")
+    CITY = "Paris"
+    URL = f"https://api.weatherapi.com/v1/current.json?key={API_KEY}&q={CITY}"
```
YuriiKindrat: it is better to place constants at the start of file

YuriiKindrat: also there is no need to use f-string for url params, use `params` param for get method

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,6 +1,40 @@
...
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    response = requests.get(
+        "https://api.weatherapi.com/v1/current.json",
+        params={
+            "key": os.environ.get("API_KEY"),
+            "q": "Paris"
```
YuriiKindrat: it seems to me, that u have the same constants ))

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,6 +1,40 @@
...
+API_KEY = os.environ.get("API_KEY")
+CITY = "Paris"
+URL = "https://api.weatherapi.com/v1/current.json"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    response = requests.get(
+        "https://api.weatherapi.com/v1/current.json",
```
YuriiKindrat: what about url ?))

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #266, opened 2023-08-08
implemented weather-api

mariiahorbova:


app/main.py
```diff
@@ -1,6 +1,32 @@
...
-    # write your code here
-    pass
+    """
+    Returns string representation of weather in Paris in following format:
+    {city}/{country} {date time} Weather: {temperature} Celsius, {text}
+    """
+
+    print(f"Performing request to Weather API to city {CITY}")
+
+    result_json = requests.get(URL + f"key={API_KEY}&q={CITY}").json()
```
YuriiKindrat: no need to use f-string for url params 
use `params` param for get method )

app/main.py
```diff
@@ -1,6 +1,32 @@
...
+
+    city = result_json["location"]["name"]
+    country = result_json["location"]["country"]
+    localtime = result_json["location"]["localtime"]
+    temp_celsius = result_json["current"]["temp_c"]
+    condition = result_json["current"]["condition"]["text"]
+
+    result_str = (f"{city}/{country} {localtime} "
+                  f"Weather: {temp_celsius} Celsius, {condition}"
+                  )
```
YuriiKindrat: bad formatting here (
also no need to create additional var here, u can palce string directly to the print func )

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #275, opened 2023-08-13
Docker Weather API

N1khto:


app/main.py
```diff
@@ -1,6 +1,23 @@
...
+URL = "https://api.weatherapi.com/v1/current.json?"
+FILTERING = "Paris"
+API_KEY = os.environ.get("API_KEY")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    result = requests.get(URL + f"q={FILTERING}&key={API_KEY}")
+    result = json.loads(result.content)
```
ArsenPidhoretskyi: You can do as follow:
result.json()

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+    if result.get("error"):
+        print(f"{result.get('error').get('message')}")
```
ArsenPidhoretskyi: If .get('error') return none you will have error
Add default argument

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+        return
+    print(f"{result.get('location').get('tz_id')} "
+          f"{result.get('current').get('last_updated')} "
+          f"Weather: {result.get('current').get('temp_c')} Celsius",
+          f"{result.get('current').get('condition').get('text')}")
```
ArsenPidhoretskyi: The same with .get
Use [] or add default

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
.

app/main.py
```diff
@@ -1,6 +1,22 @@
...
+API_KEY = os.environ.get("API_KEY")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    print(f"Performing request to Weather API for city {FILTERING}...")
+    result = (requests.get(URL + f"q={FILTERING}&key={API_KEY}")).json()
+    if result.get("error"):
+        print(f"{result.get('error').get('message', 'oops, something wrong')}")
```
ArsenPidhoretskyi:
```suggestion
        print(f"{result['error'].get('message', 'oops, something wrong')}")
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Good job!

PR #276, opened 2023-08-13
solution

OksanaAdamchuk:


app/main.py
```diff
@@ -1,6 +1,28 @@
...
+PARAMS = {"key": KEY, "q": CITY}
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    result = requests.get(URL, params=PARAMS)
+    data = result.json()
+
+    city = data["location"]["name"]
```
ArsenPidhoretskyi: Add info if error occurs. result.status_code is not 200

ArsenPidhoretskyi (changes requested):

vsmutok (approved):
Looks Good 👍

PR #279, opened 2023-08-22
Solution

olenaliuby:
https://hub.docker.com/r/oliuby/weather-app

Screenshot 2023-08-22 at 12 31 28


app/main.py
```diff
@@ -1,6 +1,27 @@
...
-    pass
+    params = {
+        "key": os.environ["API_KEY"],
+        "q": FILTERING
+    }
+    response = requests.get(URL, params=params)
+
+    if response.status_code == 200:
+        data = response.json()
+        tz_id = data["location"]["tz_id"]
```
vasylhnatiuk: I think better to use get() here and in the below code

vasylhnatiuk (changes requested):

app/main.py
olenaliuby: Thank you! This way is definitely a better option, since it won't raise an exception if the key doesn't exist.

vasylhnatiuk (approved):

PR #289, opened 2023-09-30
Solution

Munkoor:


Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+ENV PYTHONNUNBUFFERED 1
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "manage.py", "runserver", "0.0.0.0:8000"]
```
vasylhnatiuk: add newline

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+
+    response = get(base_url, params=params)
+
+    if response.status_code == 200:
+        data = response.json()
+        temperature = data["current"]["temp_c"]
+        condition = data["current"]["condition"]["text"]
+        print(
+            f"Paris/France {datetime.date.today()} "
+            f"Weather: {temperature}°C, {condition}")
```
vasylhnatiuk:
```suggestion
            f"Weather: {temperature}°C, {condition}"
        )
```

vasylhnatiuk (changes requested):

Dockerfile
vsmutok, line 12: why runserver????

vsmutok (changes requested):

Dockerfile
Munkoor: why not?

vsmutok: you `runserver` like this only in Django projects

vsmutok: and this is not a django project it will not work

Munkoor: forgot about it

vsmutok: attach a screenshot of the task result from the terminal so that it can be seen that it is running in docker

Munkoor: image

vsmutok, line 12: fix it

vsmutok (changes requested):

PR #294, opened 2023-10-12
Solved

Dimo-n:


app/main.py
```diff
@@ -1,6 +1,37 @@
...
-    pass
+    payload = {
+        "key": API_KEY,
+        "q": FILTER,
+    }
+
+    response = requests.get(URL, params=payload)
+
+    if response.status_code != 200:
+        print(f"No data: error code {response.status_code}")
```
Dimosphen1, line 21: Return in case there is not 200 response, further code should not execute after that.

app/main.py
```diff
@@ -1,6 +1,37 @@
+import requests
+from dotenv import load_dotenv
+import os
```
Dimosphen1: Order your imports in a correct way.

Dimosphen1 (changes requested):
Several changes were requested.

Dimosphen1 (approved):
LGTM, well done!

PR #296, opened 2023-10-17
Solution

OlegatorLE:


nyanplague (approved):

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+PARAMS = {
+    "key": API_KEY,
+    "q": CITY,
+}
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    res = get(url=URL, params=PARAMS)
```
Oleg-s-s: result is not so big to shorten it

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+
+    if res.status_code == 200:
```
Oleg-s-s: add some response if status not 200

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,6 +1,30 @@
...
+        weather_data = response.json()
+
+        city_datetime = weather_data["location"]["localtime"]
+        cur_temp = weather_data["current"]["temp_c"]
+        condition = weather_data["current"]["condition"]["text"]
+        print(f"Current weather in {CITY} at {city_datetime}: \n"
+              f"{cur_temp}°C \n"
+              f"{condition}")
+    else:
+        print(f"Bad request: {response.status_code}")
```
Oleg-s-s: Bad request is status code 400, so it will confuse the user if status code will be 500, 502, 402, 401 etc.
Make it more abstract, + move this block to the top to improve readability:
`
if not ...
else:
`

Oleg-s-s (changes requested):

app/main.py
Sofbog: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400

Serhii-Leonenko (approved):

PR #304, opened 2023-10-18
Solution

CaCtUs1k:


app/main.py
```diff
@@ -1,6 +1,29 @@
...
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+
+    api_key = os.environ.get("API_KEY")
+    city = "Paris"
+    url = ("https://api.weatherapi.com/v1/current.json?"
+           f"key={api_key}&q={city}")
```
denys-source: read the checklist. i think it is better to specify query params using `params=` argument

katryana (approved):

app/main.py
Oleg-s-s: Yes, and constants should be uppercased, f.e. BASE_URL, API_KEY, CITY

Oleg-s-s (changes requested):

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:3.11-slim
+LABEL maintainer="denischernish19012004@gmail.com"
+
+ENV PYTHONBUFFERED 1
```
Oleg-s-s: Its more useful to use unbuffered python in docker

Oleg-s-s (changes requested):

Oleg-s-s (approved):

PR #308, opened 2023-10-19
Done

maxkatkalov:
image

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -11,20 +11,26 @@ URL = "https://api.weatherapi.com/v1/current.json"
 
 def get_weather() -> None:
     response = requests.get(URL, params={"q": CITY, "key": KEY})
-
     response_json = response.json()
-
     weather_data = response_json.get("current")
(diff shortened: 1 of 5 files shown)


ORENYT (approved):
I would use [""] instead of get, but LGTM

app/main.py
```diff
@@ -1,6 +1,30 @@
...
+
+    weather_data = response_json.get("current")
+    location_data = response_json.get("location")
+
+    result = (
+        f"{location_data.get('name')}/"
+        f"{location_data.get('country')}"
+        f" {location_data.get('localtime')}"
+        f" Weather: {weather_data.get('temp_c')} Celsius,"
+        f" {weather_data.get('condition').get('text')}"
```
Oleg-s-s: AttributeError if condition is None

Oleg-s-s (changes requested):

Sofbog (approved):
LGTM

PR #309, opened 2023-10-19
Implemented weather API

ORENYT:
image

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -4,7 +4,7 @@ import requests
 
 API_KEY = os.environ.get("API_KEY")
 CITY = "Paris"
-URL = f"http://api.weatherapi.com/v1/current.json"
+URL = "http://api.weatherapi.com/v1/current.json"
 
 
 def get_weather() -> None:
--- a/app/main.py
+++ b/app/main.py
@@ -4,7 +4,7 @@ import requests
 
 
 def get_weather() -> None:
-    print(f"Performing request to Weather API for city Paris...")
+    print("Performing request to Weather API for city Paris...")
     api_key = os.environ.get("API_KEY")
     if api_key is None:
         raise Exception("API Key not found.")
@@ -16,8 +16,8 @@ def get_weather() -> None:
         raise Exception("Bad request. Check API key and WeatherAPI status")
 
 
-def retrieve_result(input):
-    data = input.json()
+def retrieve_result(input_response: requests.Response) -> None:
+    data = input_response.json()
     print(
         f"{data['location']['name']}/{data['location']['country']} "
         f"{data['location']['localtime']} "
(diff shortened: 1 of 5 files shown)


9rosLove (approved):

app/main.py
Oleg-s-s, line 1: use constants for url key etc

Oleg-s-s (changes requested):

app/main.py
ORENYT: describe your suggest more, please

app/main.py
```diff
@@ -1,6 +1,29 @@
+import os
+
+import requests
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    print("Performing request to Weather API for city Paris...")
+    api_key = os.environ.get("API_KEY")
```
Oleg-s-s: use constant for api_key

app/main.py
```diff
@@ -1,6 +1,29 @@
...
+    if api_key is None:
+        raise Exception("API Key not found.")
+    url = f"http://api.weatherapi.com/v1/current.json?key={api_key}&q=Paris"
```
Oleg-s-s: use constant for url and split it into base url and other stuff

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,6 +1,33 @@
...
-    # write your code here
-    pass
+    print("Performing request to Weather API for city Paris...")
+    if API_KEY is None:
+        raise Exception("API Key not found.")
+    response = requests.get(URL)
+    if response.status_code == 200:
+        retrieve_result(response)
+    else:
+        raise Exception("Bad request. Check API key and WeatherAPI status")
```
Oleg-s-s: bad request is status code 400, consider using more abstract name

app/main.py
```diff
@@ -1,6 +1,33 @@
+import os
+
+import requests
+
+API_KEY = os.environ.get("API_KEY")
+CITY = "Paris"
+URL = f"http://api.weatherapi.com/v1/current.json?key={API_KEY}&q={CITY}"
```
Oleg-s-s: consider using constant url as a base url, in this example:
`http://api.weatherapi.com/v1/current.json?`
And build a full url in `get_weather()`

Oleg-s-s (changes requested):
Looks great, but I have a few recommendations

app/main.py
ORENYT: I don't understand what do you want and why. You adviced to use url as constant, it is applied. This task does not require additional work with url, so one constant is enough. It is redundant feature.

app/main.py
ORENYT: It is console app, not a django app, do you really reccomend to use django's errors here?

Serhii-Leonenko (approved):

PR #315, opened 2023-10-19
'Solution'

grungezz:


Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+WORKDIR /app
+
+COPY requirements.txt requirements.txt
+
+RUN pip install -r requirements.txt
+
+COPY . .
+
+
+CMD ["python", "app/main.py"]
```
Serhii-Leonenko: don't forget about a blank line

app/main.py
```diff
@@ -1,6 +1,31 @@
+import os
+
+import requests
+
+
+api_key = os.environ.get("API_KEY")
```
Serhii-Leonenko: you can make this as a constant as well

app/main.py
```diff
@@ -1,6 +1,31 @@
...
+
+BASE_URL = "https://api.weatherapi.com/v1/current.json?q=Paris&key="
```
Serhii-Leonenko: It doesn't look like base url, also don't pass query parameters directly in the string, you can use` params=` attribute in the request.get() method

app/main.py
```diff
@@ -1,6 +1,31 @@
...
+
+URL = BASE_URL + api_key
+
+
 def get_weather() -> None:
     # write your code here
```
Serhii-Leonenko: redundant comment

Serhii-Leonenko (changes requested):

obohatov (approved):
LGTM

app/main.py
```diff
@@ -1,6 +1,30 @@
...
+    "q": "Paris",
+    "key": API_KEY
+}
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    print("Performing request to Weather API for city Paris...")
+    response = requests.get(BASE_URL, params=params)
```
Sofbog, line 16: its good to check if response was successful, otherwise show an error

Sofbog (changes requested):

Sofbog (approved):
LGTM

PR #319, opened 2023-10-20
Solution

ochernous:


Serhii-Leonenko (changes requested):
all checks should be passed, also don't forget about blank lines

Serhii-Leonenko (approved):

PR #324, opened 2023-10-26
Develop

TarasSavchyn:


ArsenPidhoretskyi (changes requested):
Well done!

app/main.py
```diff
@@ -1,6 +1,29 @@
...
 def get_weather() -> None:
-    # write your code here
-    pass
+    params = {"key": API_KEY, "q": FILTERING}
+    res = requests.get(URL, params=params).json()
+    if res.get("error") is not None:
+        print(res.get("error"))
+        return
+    print(
+        f"Country: {res.get('location').get('country')};\n"
```
ArsenPidhoretskyi: not good to write .get().get()
Because in this case NoneType error possible add default as dict in all places

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #338, opened 2023-11-30
Solution

VasylTurok:


app/main.py
```diff
@@ -1,6 +1,28 @@
...
-    # write your code here
-    pass
+    response = requests.get(
+        "https://api.weatherapi.com/v1/current.json",
+        params={
+            "key": API_KEY,
+            "q": "Paris"
+        }
+    )
+    weather_data = response.json()
```
ArsenPidhoretskyi: Also check that response is success(status code is equal to 2xx)

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,6 +1,36 @@
...
+    response = requests.get(
+        "https://api.weatherapi.com/v1/current.json",
+        params={
+            "key": API_KEY,
+            "q": "Paris"
+        }
+    )
+
+    # Check if the request was successful (status code 2xx)
+    if response.status_code // 100 == 2:
```
vsmutok: why you need `// 100`???

vsmutok (changes requested):

app/main.py
VasylTurok: I check if the first digit is 2

vsmutok:
```suggestion
    if response.status_code == 200:
```

vsmutok: you can write here just like that

vsmutok (changes requested):

vsmutok (approved):
Well Done 👍

PR #341, opened 2023-12-03
Solution

Vitalii-Khmura:


app/main.py
```diff
@@ -1,6 +1,31 @@
+import os
+import requests
+
+URL = ("http://api.weatherapi.com/v1/current."
+       "json?key=9ed0ea20d5ea4cf2bbd224726230212&q=Paris&aqi=yes")
```
YukihiroSM: you have a key in URL. That's bad practice. Remove the key from code. This URL should be only URL without query parameters

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1,6 +1,30 @@
...
+    weather_data = response.json()
+    localtime = weather_data["location"]["localtime"]
+    temp_c = weather_data["current"]["temp_c"]
+    condition = weather_data["current"]["condition"]["text"]
+    print("Performing request to Weather API for city Paris...")
+    print(
+        f"Paris/France {localtime} \n"
+        f"Weather: {temp_c} Celsius, \n"
+        f"{condition}"
+    )
```
EdAlekseiev: Firstly, you need to check that response status is `200`, And only after that you can show information to user

EdAlekseiev (changes requested):

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ENV PYTHONUNBUFFERED 1
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
ArsenPidhoretskyi, line 13: Add empty line

app/main.py
```diff
@@ -1,6 +1,31 @@
...
+        weather_data = response.json()
+        localtime = weather_data["location"]["localtime"]
+        temp_c = weather_data["current"]["temp_c"]
+        condition = weather_data["current"]["condition"]["text"]
+        print("Performing request to Weather API for city Paris...")
+        print(
+            f"Paris/France {localtime} \n"
+            f"Weather: {temp_c} Celsius, \n"
+            f"{condition}"
+        )
```
ArsenPidhoretskyi, line 28: But if the response is not success? Maybe also need to add some print and inform user about this

ArsenPidhoretskyi (changes requested):

EdAlekseiev (approved):
Good job

PR #346, opened 2023-12-06
Solution

prochigor:


mykolamateichuk (approved):
👍

app/main.py
```diff
@@ -1,6 +1,38 @@
+import requests
+import os
+
+from dotenv import load_dotenv
+
+load_dotenv()
+
+URL = "http://api.weatherapi.com/v1/current.json"
+FILTERING = "Paris"
+API_KEY = os.environ["API_KEY"]
```
Y-Havryliv: if the env variable is not found, your code will break

app/main.py
```diff
@@ -1,6 +1,38 @@
...
 def get_weather() -> None:
-    # write your code here
-    pass
+
+    params = {
+        "q": FILTERING,
+        "key": API_KEY
+    }
+
+    response = requests.get(URL, params=params)
```
Y-Havryliv, line 20: you forgot to print: Performing request to Weather API for city Paris...

Y-Havryliv (changes requested):
also, you were supposed to change README.md.
`You must modify this line with correct command to pull your image: COMMAND=docker pull <YOUR_DOCKER_ID/YOUR_IMAGE_NAME>.`

app/main.py
prochigor: ok, I write key in python file

Y-Havryliv (approved):

PR #347, opened 2023-12-06
Develop

vshvanska:


Shatrovskyi (approved):
Well done!

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+API_KEY = os.environ.get("WEATHER_API_KEY")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    response = requests.get(URL,
+                            params={"key": API_KEY,
+                                    "q": CITY,
+                                    "aqi": "no"})
```
LiudmylaKulinchenko: consider to create dict with params outside of method to make it more readable

LiudmylaKulinchenko (changes requested):
`You must modify this line with correct command to pull your image: COMMAND=docker pull <YOUR_DOCKER_ID/YOUR_IMAGE_NAME>.` -- fix that in your readme

.dockerignor
```diff
@@ -0,0 +1 @@
+venv
```
wQuelS: better to add `venv/` as a folder

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:3.11.4
```
wQuelS: it is almost always better to use light version to not install "full python". For example `FROM python:3.10-slim`  - but with needed version

app/main.py
```diff
@@ -1,7 +1,27 @@
-def get_weather() -> None:
-    # write your code here
-    pass
-
-
-if __name__ == "__main__":
-    get_weather()
+import os
+import requests
+from dotenv import load_dotenv
```
wQuelS: what about spaces between different group of imports?

wQuelS (changes requested):

.dockerignor
Y-Havryliv, line 1: it should be named `.dockerignore`

Y-Havryliv (approved):
GJ!

PR #350, opened 2023-12-07
Solution

Shatrovskyi:
docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM python:3.10.10
+LABEL maitainer="shatrovskyi.vlad@gmail.com"
+
+ENV PYTHONUNBUFFERED 1
+
+WORKDIR /usr/src/app
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10.10
+FROM python:3.10.10-slim
 LABEL maitainer="shatrovskyi.vlad@gmail.com"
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:3.10.10
```
RabenkoYevhenii: There was a requirement in task to make image of service as thin as possible. I think you should use here 
`FROM python:3.12.0-slim` or `FROM python:3.12.0-alpine`.

RabenkoYevhenii (changes requested):

Shatrovskyi:
docker

app/main.py
```diff
@@ -1,6 +1,39 @@
...
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    response = requests.get(
+        url=URL,
+        params={
+            "key": API_KEY,
+            "q": "Paris"
```
LiudmylaKulinchenko: make a city also a constant

app/main.py
```diff
@@ -1,6 +1,39 @@
...
+
+    if response.status_code == 200:
+        weather_data = response.json()
+        temp_c = weather_data["current"]["temp_c"]
+        temp_f = weather_data["current"]["temp_f"]
+        condition = weather_data["current"]["condition"]["text"]
+        wind_kph = weather_data["current"]["wind_kph"]
+        wind_dir = weather_data["current"]["wind_dir"]
+
+        print("Generating short weather forecast for PARIS...")
```
LiudmylaKulinchenko: don't forget to change also here

LiudmylaKulinchenko (changes requested):

Y-Havryliv (approved):

PR #353, opened 2023-12-07
Solution

YuliiaSaliuta:


SevKrok (approved):
Nice 👍

app/main.py
```diff
@@ -1,6 +1,30 @@
...
+}
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    result = requests.get(URL, params=PARAMS)
+    json_result = json.loads(result.content)
+
+    print(json_result)
```
Y-Havryliv: do you need this print?

app/main.py
```diff
@@ -1,6 +1,30 @@
...
+
+    date = json_result["location"]["localtime"]
+    temperature = json_result["current"]["temp_c"]
+    condition = json_result["current"]["condition"]["text"]
+
+    print(
+        f"Paris/France {date} Weather: {temperature} Celsius, {condition}"
```
Y-Havryliv: why did you hardcode it? It's better to take city and country from parsed result

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,6 +1,32 @@
+import os
+import json
+import requests
+
+
+URL = "https://api.weatherapi.com/v1/current.json"
+FILTERING = "Paris"
```
LiudmylaKulinchenko, line 7: as for me, better to name it city, not filtering

LiudmylaKulinchenko (approved):
Nice work! Also added you some recomendation 👇

app/main.py
LiudmylaKulinchenko: I got it, you have an example in checklist with such naming. It's totally fine 🙌

PR #357, opened 2023-12-07
Docker complete

Sel-Fisher:


Naz-iv (approved):

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+API_KEY = os.environ.get("API_KEY")
+CITY = "Paris"
+URL = "http://api.weatherapi.com/v1/current.json"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    print(f"Performing request to Weather API for city {CITY}...")
+    response = requests.get(URL + f"?key={API_KEY}&q={CITY}")
```
Y-Havryliv, line 11: I know it's not in the checklist, but it's better to create `params` variable with key and q and then do:
```python
requests.get(URL, params=params)
```

app/main.py
```diff
@@ -1,6 +1,23 @@
...
+    weather_data = response.json()
+    if weather_data:
```
Y-Havryliv, line 13: also, you can receive not 200 status code, and then your code will break)

Y-Havryliv (approved):

PR #375, opened 2023-12-19
Solution

markus2812:
Також посилання на DokerHub
https://hub.docker.com/r/markus9869/my-weather-app

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -20,11 +20,11 @@ def get_weather() -> None:
     response = requests.get(BASE_URL, params=params)
     weather_data = response.json()
 
-    city_name = weather_data["location"]["name"]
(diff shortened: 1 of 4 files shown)


app/main.py
```diff
@@ -1,6 +1,37 @@
...
+
+    params = {
+        "key": api_key,
+        "q": CITY,
+    }
+
+    response = requests.get(BASE_URL, params=params)
+    weather_data = response.json()
+
+    city_name = weather_data["location"]["name"]
```
vasylhnatiuk: use get(). what if dict doesn't have this key?

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #407, opened 2024-03-06
Solved

Dmitriy-Poplinski:


Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "./app/main.py"]
+
```
viktoria-rybenchuk: remove redundant blank line

app/main.py
```diff
@@ -1,7 +1,28 @@
...
-    # write your code here
-    pass
+from requests import get
+import os
+
+from constants import BASE_URL
+
+
+def get_weather(city: str) -> None:
+    api_key = os.environ.get("WEATHER_API_KEY")
```
viktoria-rybenchuk: define. `api_key` as constant as well

app/main.py
```diff
@@ -1,7 +1,28 @@
...
+
+    result = get(BASE_URL, {
+        "key": api_key,
+        "q": city,
```
viktoria-rybenchuk, line 10: the same here

viktoria-rybenchuk (changes requested):

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ENV PYTHONBUFFERED 1
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "./app/main.py"]
```
Dimosphen1, line 13: There is no need to keep several app folders in docker container since you set that your working directory is `app/` already. You could omit using app directory in your project, thus shortening the call to a file
```suggestion
CMD ["python", "main.py"]
```

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #413, opened 2024-03-12
Solution

Yevheniia-Ilchenko:


app/main.py
```diff
@@ -1,6 +1,16 @@
...
+URL = "https://api.weatherapi.com/v1/current.json"
+CITY = "Paris"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    params = {"key": API_KEY, "q": CITY}
+    res = requests.get(url=URL, params=params).json()
+    print(res)
```
EdAlekseiev: You should check http code of response. If code is not 200, you should show some notification about it to user

EdAlekseiev (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #417, opened 2024-03-18
Solution

ArturPoltser:


app/main.py
```diff
@@ -1,6 +1,26 @@
...
+CITY = "Paris"
+URL = f"http://api.weatherapi.com/v1/current.json?key={API_KEY}&q={CITY}"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    result = requests.get(URL)
+    result.raise_for_status()
+    data = result.json()
```
aLEKS-e3: you can rewrite `data = result.json()["current"]` so it won't repeat and will be more readable

aLEKS-e3 (approved):

app/main.py
```diff
@@ -1,6 +1,26 @@
+import os
+from datetime import datetime
+
+import requests
+
+API_KEY = os.environ.get("API_KEY")
+CITY = "Paris"
+URL = f"http://api.weatherapi.com/v1/current.json?key={API_KEY}&q={CITY}"
```
Serhii-Leonenko:
```suggestion
URL = f"http://api.weatherapi.com/v1/current.json
```

use const without external constants here

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    result = requests.get(URL)
```
Serhii-Leonenko: add query params city and api key here

image

app/main.py
```diff
@@ -1,6 +1,26 @@
...
+    result.raise_for_status()
```
Serhii-Leonenko: you can add an exception handler and print smth if some error occurred

Serhii-Leonenko (changes requested):

Y-Havryliv (approved):
Well done!

PR #424, opened 2024-03-23
Weather API Solution

barabarinov:


kabust (approved):

app/main.py
```diff
@@ -1,6 +1,36 @@
...
+            f"City: {response_location.get('name')}\n"
+            f"Local datetime: {response_location.get('localtime')}\n"
+            f"Cloud: {response_current.get('cloud')}\n"
+            f"{response_current.get('condition').get('text')}\n"
+            f"Humidity: {response_current.get('humidity')}\n"
+            f"Temperature °C: {response_current.get('temp_c')}\n"
+            f"Feels like °C: {response_current.get('feelslike_c')}\n"
+        )
+    else:
+        print("Oops! Something went wrong! :)")
```
MNDonut: it's better to return response error

MNDonut (changes requested):

.dockerignore
```diff
@@ -0,0 +1,5 @@
+.python-version
+Dockerfile
```
MNDonut: let's don't ignore dockerfile

app/main.py
```diff
@@ -1,6 +1,38 @@
...
+BASE_URL = "https://api.weatherapi.com/v1/current.json"
+API_KEY = os.getenv("API_KEY")
+City = "Paris"
+LANG = "uk"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    response = requests.get(f"{BASE_URL}?key={API_KEY}&q={City}&lang={LANG}")
```
Serhii-Leonenko: you can use `params=` keyword argument for query params

Serhii-Leonenko (approved):

app/main.py
barabarinov: I did "params" before but changed it because I thought this approach is better!

PR #427, opened 2024-03-25
Solution1

projects2312:


app/main.py
```diff
@@ -1,6 +1,30 @@
...
+
+    result = requests.get(URL, {"key": API_KEY, "q": FILTERING})
+
+    if result.status_code == 200:
+        print("Performing request to weather API for city Kyiv...")
+        result = result.json()
+        location = result["location"]
+        weather = result["current"]
+        data = (
+            f"{location['name']}/{location['country']} "
```
juliastetsko, line 20: better to use `"name"`

juliastetsko (approved):
looks good

.dockerignore
```diff
@@ -0,0 +1 @@
+.venv
```
MNDonut, line 1: do you want result.png for example be present in docker image? let's ignore all redundant files

MNDonut (approved):

PR #428, opened 2024-03-25
Implement py-docker-weather-api and add to docker

juliastetsko:


app/main.py
```diff
@@ -1,6 +1,26 @@
...
+
+    if response.status_code != requests.codes.ok:
+        print(f"Request failed. Status code: {response.status_code}. "
+              f"Message: {response.text}")
+        return
+    data = response.json()
+    print(f"{data["location"]["name"]}/{data["location"]["country"]} "
+          f"{data["location"]["localtime"]} "
+          f"Weather: {data["current"]["temp_c"]} Celsius, "
+          f"{data["current"]["condition"]["text"]}")
```
mdubyna: u can use variables to describe each parameter, for example: city = data["location"]["name"]

mdubyna (approved):
GJ!

app/main.py
juliastetsko: I thought it would not be very obvious

.dockerignore
```diff
@@ -0,0 +1,2 @@
+.venv
+Dockerfile
```
MNDonut: empty line at the end of file + ignore more files that aren't needed to build or app (like readme, images(there is result.png in the repo), and so on)

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+ENV PYTHONUNBUFFERED 1
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
MNDonut: empty line at the end of file

app/main.py
```diff
@@ -1,6 +1,26 @@
...
-    # write your code here
-    pass
+    print("Performing request to Weather API for city Paris...")
+    response = requests.get(URL,
+                            params={"key": API_KEY, "q": FILTERING})
+
+    if response.status_code != requests.codes.ok:
+        print(f"Request failed. Status code: {response.status_code}. "
+              f"Message: {response.text}")
+        return
```
MNDonut, line 18: in general we use if status_code != 200, because it's GET endpoint that returns either 200 code(success) or 4xx 5xx codes(fail), but this approach is ok as well

MNDonut (changes requested):

Y-Havryliv (approved):
gj!

PR #430, opened 2024-03-25
Solution

AntonSkyrda:


NataliiaKorzh (approved):
GJ!

.dockerignore
```diff
@@ -0,0 +1,3 @@
+.git
+__pycache__
+venv
```
MNDonut, line 3: ignore more files and images, readme, etc that is not need to run container in docker

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+ENV API_KEY 6aee2b17b8a842018aa122043242503
+
+WORKDIR docker_app/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
MNDonut: empty line at the end of file

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.11.0-slim
+LABEL maintainer="mr.skyrda@gmail.com"
+
+ENV PYTHONUNBUFFERED 1
+ENV API_KEY 6aee2b17b8a842018aa122043242503
```
MNDonut: api key shouldn't be added explicitly in Dockerfile

app/main.py
```diff
@@ -1,6 +1,33 @@
...
+
+API_KEY = os.environ.get("API_KEY")
+URL = "http://api.weatherapi.com/v1/current.json?"
+
+
+def get_weather(city: str = "Paris") -> requests.Response:
+    print(f"Requesting {city} weather from Weather API...")
+    request = requests.get(
+        URL + f"key={API_KEY}&q={city}&aqi=no"
+    ).json()
```
MNDonut: use params= parameter in requests.get method when you want to add query params instead of manually constructing the url

app/main.py
```diff
@@ -1,6 +1,33 @@
...
+
+    if "error" in request:
+        print(request["error"]["message"])
+        return request
```
MNDonut: it's better to check if status_code != 200: raise error

MNDonut (changes requested):

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+ENV API_KEY
+
+WORKDIR docker_app/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["sh", "-c", "python -m dotenv -f .env python app/main.py"]
```
Serhii-Leonenko, line 14: api key should be taken from the command 

image

Serhii-Leonenko: no need to use dotenv

app/main.py
```diff
@@ -1,6 +1,31 @@
...
+URL = "http://api.weatherapi.com/v1/current.json?"
+
+
+def get_weather(city: str = "Paris") -> requests.Response:
+    print(f"Requesting {city} weather from Weather API...")
+    params = {"key": API_KEY, "q": city, "aqi": "no"}
+    response = requests.get(URL, params=params)
+
+    if response.status_code != 200:
+        response.raise_for_status()
```
Serhii-Leonenko, line 14: you can use raise for status without if condition

Serhii-Leonenko (changes requested):
image

PR #432, opened 2024-03-25
Solution

MaxOperchuk:


oksankakot (approved):
What an incredible job!

.dockerignore
```diff
@@ -0,0 +1,2 @@
+.venv
+Dockerfile
```
MNDonut, line 2: result.png?

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+import os
+
+
+URL = "http://api.weatherapi.com/v1/current.json?"
+FILTERING = "Kyiv"
+
+
+def get_weather() -> str:
+
+    res = requests.get(f"{URL}key={os.environ.get('API_KEY')}&q={FILTERING}")
```
MNDonut: use params= parameter of requests.get() method instead of constructing url with query params

MNDonut (changes requested):

MNDonut (approved):

PR #433, opened 2024-03-25
Solution

oksankakot:


Dobryk12 (approved):
Oh my God, so good

.dockerignore
```diff
@@ -0,0 +1 @@
+.venv
```
MNDonut, line 1: ignore result.png and other files as well

.env.sample
```diff
@@ -0,0 +1 @@
+API_KEY=fc1df8d1325d4868854122734242503
```
MNDonut: env file shouldn't be present in github

app/main.py
```diff
@@ -1,6 +1,53 @@
...
+            "q": CITY,
+            "units": "metric",
+            "lang": LANGUAGE
+        }
+    )
+    data = res.json()
+
+    if "error" in data:
+        print("Error:", data["error"]["message"])
+    else:
```
MNDonut: it's better to check status_code and indeed print error

MNDonut (changes requested):

.env.sample
oksankakot: removed .env.sample

app/main.py
oksankakot: Fixed that

.dockerignore
oksankakot: Fixed

.env.sample
Y-Havryliv: Ihor talks about `.env`, not `.env.sample`. Just don't store the actual keys in `.env.sample` and it's ok.

Y-Havryliv (approved):

PR #435, opened 2024-03-25
Develop

Dobryk12:


app/main.py
```diff
@@ -1,6 +1,29 @@
...
+
+# os.environ["API_KEY"] = "7c50588873ce40198ea125550242503"
+API_KEY = os.environ.get("API_KEY")
+URL = "https://api.weatherapi.com/v1/current.json"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    params = {"key": API_KEY, "q": "Paris"}
```
msymonovych: I believe you need to create global variable for "Paris" as well

msymonovych (approved):

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+LABEL maintainer="example@example.com"
+
+ENV PYTHOUNNBUFFERED 1
+
+WORKDIR weather-api/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
```
MNDonut, line 11: You should have env variable. Check the checklist

app/main.py
```diff
@@ -1,6 +1,29 @@
+import os
+
+import requests
+
+# os.environ["API_KEY"] = "7c50588873ce40198ea125550242503"
+API_KEY = os.environ.get("API_KEY")
+URL = "https://api.weatherapi.com/v1/current.json"
```
MNDonut: remove that comment with api key

app/main.py
MNDonut: Paris string should be some constant, not hardcoded string

app/main.py
```diff
@@ -1,6 +1,29 @@
...
+    temperature = data["current"]["temp_c"]
+    weather = data["current"]["condition"]["text"]
+
+    print(
+        f"Country: {country}\n"
+        f"City: Paris\n"
+        f"Time: {time}\n"
+        f"Temperature: {temperature}\n"
+        f"Weather: {weather}"
+    )
```
MNDonut: what if you get 400 or 401 or 500 error from that API? You won't get data and everything will fail. Let's handle this case

.dockerignore
```diff
@@ -0,0 +1,10 @@
+.venv
+venv
+.idea*
+.git*
+.gitignore
+.flake8
+*.md
+*.png
+Dockerfile
+.dockerignore
```
MNDonut: let's don't ignore these files

MNDonut (changes requested):

app/main.py
```diff
@@ -1,6 +1,32 @@
...
+        weather = data["current"]["condition"]["text"]
+
+        print(
+            f"Country: {country}\n"
+            f"City: Paris\n"
+            f"Time: {time}\n"
+            f"Temperature: {temperature}\n"
+            f"Weather: {weather}"
+        )
+    except requests.exceptions.RequestException as e:
```
Y-Havryliv, line 29: there could be other errors like timeouts, error HTTP status codes, and so on...
think about how to handle that

Y-Havryliv (changes requested):

MNDonut (approved):

PR #436, opened 2024-03-25
Solution

msymonovych:


tskopenko (changes requested):
Please add the command at the Readme file, coz rn its not testable at all

.dockerignore
```diff
@@ -0,0 +1 @@
+.venv
```
MNDonut, line 1: ignore more files that are not needed for running the app(like images(result.png for example), readme, dockerfile, etc)

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+FILTERING = "Paris"
+URL = "http://api.weatherapi.com/v1/current.json"
+API_KEY = os.getenv("API_KEY", None)
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    print(f"Performing request to Weather API for city {FILTERING}")
+    response = requests.get(URL + f"?q={FILTERING}&key={API_KEY}").json()
```
MNDonut: use params parameter in .get() method instead of constructing the url manually

app/main.py
```diff
@@ -1,6 +1,24 @@
...
+    location = (f"{response['location'].get('name')}/"
+                + response["location"].get("country"))
+    localtime = response["location"].get("localtime")
+    temperature = response["current"].get("temp_c")
+    condition = response["current"]["condition"].get("text")
+
```
MNDonut: what if API returns error, thus you won't get data in the correct format? Take it into account and handle it

MNDonut (changes requested):

.dockerignore
```diff
@@ -0,0 +1,9 @@
...
+*.png
+*.md
+Dockerfile
+.git
+.github
+.gitignore
+.idea
+.flake8
```
Serhii-Leonenko: don't forget about a blank line

app/main.py
```diff
@@ -1,6 +1,30 @@
...
+        localtime = response["location"].get("localtime")
+        temperature = response["current"].get("temp_c")
+        condition = response["current"]["condition"].get("text")
+
+        print(
+            f"{location} {localtime} Weather: "
+            f"{temperature} Celsius, {condition}"
+        )
+    except Exception:
+        print("Error! Something went wrong!")
```
Serhii-Leonenko: will an exception be raised if our request is processed but the response status is 403 or smth else?
use `raise_for_status` https://3.python-requests.org/user/quickstart/#response-status-codes or check status codes

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):

PR #438, opened 2024-03-25
Task Solution

shaiduchyk:


Serejka21 (approved):
Looking good, well done

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+LABEL maintaner="jaegarn141@gmail.com"
+
+ENV PYTHOUNNBUFFERED 1
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "-m", "app.main"]
```
MNDonut: empty line

.dockerignore
```diff
@@ -0,0 +1 @@
+.venv
```
MNDonut, line 1: do you want for example result.png be present in docker image? let's ignore all redundant files

app/main.py
```diff
@@ -1,7 +1,17 @@
...
+
+
+def get_weather() -> str:
+    payload = {"key": API_KEY, "q": CITY}
+    get_weather_url = BASE_URL + "current.json"
+    data = requests.get(get_weather_url, params=payload).json()
+    return (f"{data['location']['name']}/{data['location']['country']} "
+            f"{data['location']['localtime']} "
+            f"Weather: {str(data['current']['temp_c'])} Celsius, "
+            f"{data['current']['condition']['text']}")
```
MNDonut: what if API returns 400 401 500 or other status code that returns errors, not data as you expect here? Let's handle this case as well

MNDonut: Also this function should print the result, not return

MNDonut (changes requested):

Serhii-Leonenko (approved):

PR #439, opened 2024-03-25
Solution

Serejka21:


barabarinov (approved):
GJ! :)

app/main.py
```diff
@@ -1,6 +1,25 @@
...
-    # write your code here
-    pass
+    response = requests.get(URL + METHOD + f"?key={KEY}&q={FILTER}")
+    if response.status_code == 200:
+        data = response.json()
+        print(f"Current location: {data['location']['name']}\n"
+              f"Time: {data['location']['localtime']}\n"
+              f"Temperature: {data['current']['temp_c']}")
+    else:
+        print(f"{response.status_code}")
```
MNDonut: it's better to return response error instead of status code

MNDonut: Also I don't think you want result.png file and others to be added in docker image, let's ignore it

MNDonut (changes requested):

.dockerignore
```diff
@@ -0,0 +1,3 @@
+venv
+.github
+.png
```
Serhii-Leonenko: blank line

app/main.py
```diff
@@ -1,6 +1,25 @@
...
+
+FILTER = "Kharkiv"
+
+METHOD = "current.json"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    response = requests.get(URL + METHOD + f"?key={KEY}&q={FILTER}")
```
Serhii-Leonenko: query params better to send with keyword argument `params=` and use interpolation instead of concatenation

Serhii-Leonenko (approved):

PR #441, opened 2024-03-26
Solution

JohnnyWalker010:


PythonZem (approved):
good clean clean code

set
Y-Havryliv, line 1: delete this file

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ENV PYTHONUNBUFFERED 1
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
Y-Havryliv: add a blank line

docker
Y-Havryliv, line 1: delete this file

app/main.py
```diff
@@ -1,6 +1,40 @@
...
+API_KEY = os.getenv("API_KEY")
+URL = "http://api.weatherapi.com/v1/current.json"
+CITY = "Paris"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    if not API_KEY:
+        raise ValueError("No API key provided for Weather API")
```
Y-Havryliv: probably it's better to raise some custom error `MissingAPIKeyError` or smth like this

Y-Havryliv (changes requested):

.dockerignore
```diff
@@ -0,0 +1,2 @@
+venv
+Dockerfile
```
Serhii-Leonenko, line 2: you can add more files to exclude, like gitignore, readme, idea, etc

app/main.py
```diff
@@ -1,6 +1,47 @@
...
+API_KEY = os.getenv("API_KEY")
+URL = "http://api.weatherapi.com/v1/current.json"
+CITY = "Paris"
+
+
+class MissingAPIKeyError(Exception):
+
+    def __init__(self, message: str) -> None:
+        super().__init__(message)
+        self.message = message
```
Serhii-Leonenko: do you really need to override this init here?

Serhii-Leonenko: also better to move this exception to a separate file

app/main.py
```diff
@@ -1,6 +1,47 @@
...
+        "location", {}).get("localtime", "Unknown time")
+    temp_c = weather_data.get(
+        "current", {}).get("temp_c", "value not correct")
+    condition_text = weather_data.get(
+        "current", {}).get("condition", {}).get("text", "value not correct")
+
+    print(
+        f"Weather in {city},{country} is {condition_text} at {local_time}."
+        f"Air temperature: {temp_c} degrees Celsius"
+    )
```
Serhii-Leonenko: it's better to handle unsuccessful response, use `raise_for_status` https://3.python-requests.org/user/quickstart/#response-status-codes or check status codes of the response

requirements.txt
```diff
@@ -3,3 +3,5 @@ flake8-annotations==2.9.1
 flake8-quotes==3.3.1
 flake8-variables-names==0.0.5
 pep8-naming==0.13.2
+
+requests~=2.31.0
```
Serhii-Leonenko: redundant line between rows, a blank line at the end of the file is absent

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -2,18 +2,13 @@
...
-class MissingAPIKeyError(Exception):
-
-    def __init__(self, message: str) -> None:
-        super().__init__(message)
-        self.message = message
-
-
 def get_weather() -> None:
     if not API_KEY:
         raise MissingAPIKeyError("No API key provided for Weather API")
```
MNDonut: Yaroslav meant that requests.get(URL, params=response_params(by the way, it should be request_params because you make a request)) may return 400 error and response will differ from what you expect. Like this
{
    "error": 'text'
}
or
{
     "errors": ['error 1', 'error2']
}
You don't know what kind of response you'll receive when get 400, 401, 500, etc errors, thus you should catch these errors. Like if status code isn't 200 -> print error. So check status code or catch error or use what Serhii recommended

MNDonut (changes requested):

MNDonut (approved):
Perfect!

PR #442, opened 2024-03-26
solution

PythonZem:


VladKyianytsia (approved):

app/main.py
```diff
@@ -1,6 +1,52 @@
...
+            f"City: {city}\n"
+            f"Local Time: {local_time}\n"
+            f"\nWeather\n"
+            f"Last Updated: {last_updated}\n"
+            f"Condition: {weather_condition}\n"
+            f"Temp°C: {temp_c} (feels like {feelslike_c})\n"
+            f"Wind Speed: {wind_speed} kph\n"
+        )
+    else:
+        raise ValueError(
```
Y-Havryliv: A `ValueError` is typically used to indicate that a function received an argument of the correct type but with an invalid value.
And it's not the best option to raise, think of smth better

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
Y-Havryliv: add a blank line

Y-Havryliv (changes requested):

.dockerignore
```diff
@@ -0,0 +1,3 @@
+.venv
+__pycache__
+README.md
```
Serhii-Leonenko, line 3: you can add more files to exclude

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.12-alpine
+LABEL maintainer="pythonzem@gmail.com"
+
+ENV PYTHONUNNBUFFERED 1
```
Serhii-Leonenko, line 4:
```suggestion
ENV PYTHONUNBUFFERED 1
```

Serhii-Leonenko (approved):

PR #445, opened 2024-03-26
Solution

isikach:


motivated2 (changes requested):
You need add command to pull from docker hub:
Screenshot_1

isikach:
> You need add command to pull from docker hub: Screenshot_1

app/main.py
```diff
@@ -1,6 +1,17 @@
+import os
+import requests
+
+
+API_KEY = os.getenv("API_KEY")
+CITY = "Paris"
+URL = f"http://api.weatherapi.com/v1/current.json?key={API_KEY}&q={CITY}"
```
Serhii-Leonenko: better to pass query params in the requests with `params=` keyword argument 

image

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    response = requests.get(URL)
+    data = response.json()
+    celsius = data["current"]["temp_c"]
+    print(f"In {CITY}: {celsius} now")
```
Serhii-Leonenko: what happens if the response status is not 200? you should handle this

also use prints according to the example 
image

Serhii-Leonenko (changes requested):
Fix your mate's comment

app/main.py
isikach: cool point

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+        county_name = data["location"]["country"]
+        local_time = data["location"]["localtime"]
+        celsius = data["current"]["temp_c"]
+        condition = data["current"]["condition"]["text"]
+        print(
+            f"In {CITY}/{county_name} {local_time} "
+            f"Weather {celsius} Celsius, {condition}"
+        )
+    except requests.exceptions.ConnectionError:
+        print("Connection")
```
MNDonut, line 25: This print should describe what happened

MNDonut (approved):

PR #449, opened 2024-03-31
Solution

kstorozhenko:


app/main.py
```diff
@@ -1,6 +1,34 @@
...
+BASE_URL = "https://api.weatherapi.com/v1/current.json"
+CITY = "Paris"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    if not API_KEY:
+        raise ValueError("API_KEY environment variable not set")
+    print("Getting weather for city Paris")
```
MNDonut: 1 blank line

app/main.py
```diff
@@ -1,6 +1,34 @@
...
+    url = f"{BASE_URL}?key={API_KEY}&q={CITY}&aqi=no"
+
+    response = requests.get(url)
```
MNDonut: Use params= parameter instead of constructing the url manually
https://requests.readthedocs.io/en/latest/user/quickstart/#passing-parameters-in-urls

app/main.py
```diff
@@ -1,6 +1,34 @@
...
+
+    if response.status_code == 200:
+        weather_data = response.json()
```
MNDonut, line 28: When weather api returns failed response, it's better to display this error so that you know the reason and if you need to fix it or it's an error on weather api side

MNDonut (changes requested):

Y-Havryliv (approved):
gj!

PR #456, opened 2024-04-18
Develop

nicksetrakov:


app/main.py
```diff
@@ -1,6 +1,28 @@
+import os
+
+import requests
+
+
+API_KEY = os.environ.get("API_KEY")
+URL = (
+    f"https://api.weatherapi.com/v1/current.json?key={API_KEY}&q=Paris&aqi=no"
+)
```
wQuelS, line 9: I'd suggest adding `city` and `API_Key` via optional `params` in your `get()` function to make in a more dynamic approach

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM python:3.11-alpine3.18
```
wQuelS, line 1: alpine is fine, and it is pretty thin. However, the task asks to use an image as thin as possible. 
`slim` version should be slightly thiner

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+        print(
+            f"{weather['location']['name']}/{weather['location']['country']}",
+            weather["current"]["last_updated"],
+            "Weather:",
+            weather["current"]["temp_c"],
+            "Celsius,",
+            weather["current"]["condition"]["text"]
+        )
+    else:
+        weather.raise_for_status()
```
wQuelS: Going with `if not ... == 200:` and raising an error at the top, would give the possibility to get rid of else and tabs for code that in a `status_code  == 200` part

wQuelS (changes requested):

Dockerfile
nicksetrakov: I don't see that slim is thinner than alpine
image
image

app/main.py
nicksetrakov: I understand what you mean. but we have the task, where we get weather for Paris and get API KEY from the environment

fsocie7y (approved):
good job}

PR #471, opened 2024-05-21
Solution

loplicat:


app/main.py
```diff
@@ -1,6 +1,43 @@
...
+    response = requests.get(URL, params=payload)
+    response_data = json.loads(response.content)
+
+    if response.status_code == 200:
+        print(get_str_from_dict(response_data))
+    else:
+        print(
+            f"Status code is {response.status_code}. "
+            f"Error: {response_data.get('error')}"
+        )
```
viktoria-rybenchuk: Using `try-except` allows you to handle potential exceptions that may arise, such as network errors, invalid responses, or other unexpected issues

viktoria-rybenchuk (changes requested):

app/main.py
loplicat: Thanks for the advice :) I've already replaced if statement with try-except

app/main.py
```diff
@@ -1,6 +1,42 @@
...
+    print(f"Performing request to Weather API for city {FILTERING}...")
+
+    payload = {
+        "q": FILTERING,
+        "key": API_KEY
+    }
+    response = requests.get(URL, params=payload)
+    response_data = json.loads(response.content)
+
+    try:
```
pavlopro, line 31: Your previous code also make sence
If the response code is not 200, give user info whats goes wrong
You can combine it with Viktorias suggestion

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #475, opened 2024-05-25
Docker weather API

dissom:


vsmutok (changes requested):
Tests Failed!

app/main.py
```diff
@@ -1,6 +1,30 @@
...
+        data = response.json()
+        location_name = data["location"]["name"]
+        location_country = data["location"]["country"]
+        localtime = data["location"]["localtime"]
+        temperature = data["current"]["temp_c"]
+        weather = data["current"]["condition"]["text"]
+        print(
+            f"{location_name}/{location_country} {localtime} "
+            f"Weather: {temperature} Celsiush {weather}"
+        )
```
vsmutok, line 27: what if the status is not equal to 200? how to understand that something went wrong? 

add some message for this case

.dockerignore
```diff
@@ -0,0 +1,2 @@
+venv
+Dockerfile
```
vsmutok, line 2: add other files that you should not include

vsmutok (changes requested):
image

vsmutok (changes requested):
.

vsmutok (approved):
great!

PR #481, opened 2024-06-02
Docker

mikhailyemets:


.dockerignore
```diff
@@ -0,0 +1,12 @@
...
+.vscode/
+*.iml
+.DS_Store
+.pytest_cache/
+**__pycache__/
+.git*
+.flake8
+.dockerignore
+Dockerfile
+result.png
```
YuriiKindrat: no blank line at the end of file

app/main.py
```diff
@@ -1,6 +1,26 @@
...
 def get_weather() -> None:
-    # write your code here
-    pass
+    if not API_KEY:
+        raise ValueError(
+            "No API key provided.Set the WEATHER_API_KEY environment variable."
+        )
+
+    url = f"{URL}/current.json?key={API_KEY}&q={FILTERING}"
+    response = requests.get(url)
```
YuriiKindrat: it is better to use <params> parametr instead of f-string here

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #486, opened 2024-06-03
Develop

vitalinamal:


app/main.py
```diff
@@ -1,6 +1,23 @@
...
+if not API_KEY:
+    raise ValueError("No API key found. "
+                     "Please set the API_KEY environment variable.")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    print(f"Performing request to Weather API for city {FILTERING}...")
+    result = requests.get(URL + f"q={FILTERING}&key={API_KEY}").json()
```
YuriiKindrat: it is better to use 'params' parametr instead of f-string here

YuriiKindrat (changes requested):
image
fix that )

fsocie7y (approved):
well done ^_^

PR #487, opened 2024-06-04
Develop

alonasorochynska:
While testing the insertion of different cities, I forgot that the target city was Paris, and in my code it is London. Hopefully this isn't a problem and I won't have to create a new Docker repository because of this :) Please kindly check it out: https://hub.docker.com/r/alonasorochynska/weather-api


.dockerignore
```diff
@@ -0,0 +1 @@
+.venv
```
YuriiKindrat, line 1: no blank line at the end of file

YuriiKindrat (changes requested):
image
fix that)

fsocie7y (approved):
well done ^_^

PR #491, opened 2024-06-04
Solution

Nyxan:
Знімок екрана 2024-06-04 160915

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,7 +1,25 @@
-def get_weather() -> None:
-    # write your code here
-    pass
+import os
+import requests
 
+
+def get_weather() -> None:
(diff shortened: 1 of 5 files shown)


app/main.py
```diff
@@ -1,6 +1,25 @@
+import os
+import requests
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    api_key = os.getenv("API_KEY")
+    url = "http://api.weatherapi.com/v1/current.json"
+    filtering = "Paris"
```
YuriiKindrat: it is better to write this vars as constants

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #492, opened 2024-06-04
Solution

DmytroHlazyrin:


.dockerignore
```diff
@@ -0,0 +1,10 @@
+venv
+Dockerfile
+.dockerignore
+.flake8
+.idea
+.github
+.gitignore
+checklist.md
+README.md
+result.png
```
fsocie7y: add blank line to end of file

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+ENV PYTHONUNBUFFERED=1
+
+WORKDIR /app
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
fsocie7y: here also

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,6 +1,27 @@
+import requests
+from decouple import config
+
+
+API_KEY = config("API_KEY")
+CITY_NAME = "Paris"
+URL = (f"https://api.weatherapi.com/v1/current.json"
+       f"?key={API_KEY}&q={CITY_NAME}&aqi=no")
```
YuriiKindrat: it is bad practice to use f-string for url params, use `params` param in requests.get method instead

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Looks good 🚀

PR #496, opened 2024-06-05
Develop

VitalyBashkiser:


VitalyBashkiser:
https://hub.docker.com/repository/docker/user900002/docker-weather-api/general

app/main.py
```diff
@@ -1,7 +1,22 @@
-def get_weather() -> None:
-    # write your code here
-    pass
+import os
+import requests
+
+
+def get_weather(api_key: str) -> str:
+    url = f"http://api.weatherapi.com/v1/current.json?key={api_key}&q=Paris"
```
YuriiKindrat: it is bad practice to pass url params with f-string, use `params` param in requests.get method instead

app/main.py
```diff
@@ -1,7 +1,22 @@
...
+        temp_c = data["current"]["temp_c"]
+        condition = data["current"]["condition"]["text"]
+        return f"Weather report for {location}: {temp_c}°C, {condition}"
+    else:
+        return "Failed to get weather data"
 
 
 if __name__ == "__main__":
-    get_weather()
+    api_key = os.getenv("API_KEY")
```
YuriiKindrat: it is better to move api key to constants, the same is for city

YuriiKindrat (changes requested):

YuriiKindrat (changes requested):
image
u should modify this line in readme.md file

fsocie7y (approved):
well done ^_^

PR #512, opened 2024-06-25
Solution

AlexRozv:


app/main.py
```diff
@@ -1,6 +1,20 @@
...
+
+URL = "http://api.weatherapi.com/v1/forecast.json?"
+FILTERING = "Paris"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    api_key = os.environ["API_KEY"]
+    result = requests.get(URL + f"key={api_key}&q={FILTERING}")
```
LiudmylaKulinchenko: Set parameters through params attribute

app/main.py
```diff
@@ -1,6 +1,20 @@
...
+import requests
+
+URL = "http://api.weatherapi.com/v1/forecast.json?"
+FILTERING = "Paris"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    api_key = os.environ["API_KEY"]
```
LiudmylaKulinchenko: Looks like api_key could also be constant

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good 💪

PR #524, opened 2024-07-18
Solution

KwintDimon:


vsmutok (changes requested):
you have attached only two empty files

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ENV PYTHOUNBUFFERED 1
+
+WORKDIR app/
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
pavlopro: add blank line

app/main.py
```diff
@@ -1,6 +1,33 @@
...
+            "q": CITY,
+        }
+    )
+
+    if response.status_code != 200:
+        print(f"Can't get data from Weather API. "
+              f"Status code: {response.status_code}")
+
+    data = response.json()
+    print(
```
pavlopro: Your program will fail if the status code is not 200. It would be better to retrieve the data within a block of code that checks if the status code is 200.

pavlopro (changes requested):

app/main.py
```diff
@@ -1,6 +1,33 @@
...
-    pass
+    response = requests.get(
+        URL,
+        params={
+            "key": API_KEY,
+            "q": CITY,
+        }
+    )
+
+    if response:
```
pavlopro: Check for status code 200

pavlopro (changes requested):

app/main.py
KwintDimon: Request object allows to check status code 200 in this way

pavlopro: It's better to explicitly check the status code to ensure the response was successful. This makes your code more clear and robust.

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #527, opened 2024-07-22
"Solution"

romanshupak:


EdAlekseiev (changes requested):
Your get_weather is empty. You should add your solution

app/main.py
```diff
@@ -1,6 +1,36 @@
...
+CITY = "Paris"
+API_KEY = os.environ.get("API_KEY")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    url = f"{URL}?key={API_KEY}&q={CITY}"
+    response = requests.get(url).json()
+    print(response)
```
vasylhnatiuk: delete redundandt print()

app/main.py
```diff
@@ -1,6 +1,36 @@
...
+URL = "https://api.weatherapi.com/v1/current.json"
+CITY = "Paris"
+API_KEY = os.environ.get("API_KEY")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    url = f"{URL}?key={API_KEY}&q={CITY}"
+    response = requests.get(url).json()
```
vasylhnatiuk: and you should check if response status 200 and after that you can get data

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #540, opened 2024-08-03
Solution

BohdanKuzik:


app/main.py
```diff
@@ -1,7 +1,29 @@
...
+        weather_data = response.json()
+        print(
+            f"{weather_data['location']['name']}/"
+            f"{weather_data['location']['country']} "
+            f"{weather_data['location']['localtime_epoch']} "
+            f"Temperature: {weather_data['current']['temp_c']} "
+            f"Celsius, {weather_data['current']['condition']['text']}"
+        )
+    else:
+        print(f"Failed to get weather data: {response.status_code}")
```

Using a try block helps make your code more resilient and user-friendly by managing exceptions and providing informative error messages.

Arsen-hrynevych (changes requested):

BohdanKuzik:
link to the image

https://hub.docker.com/r/bohdankuzik/docker-weather-api

Arsen-hrynevych (approved):
Thanks

PR #541, opened 2024-08-03
py-docker-weather-api v1

BornToLivee:


app/main.py
```diff
@@ -1,6 +1,21 @@
...
 def get_weather() -> None:
-    # write your code here
-    pass
+    response = requests.get(f"{BASE_URL}?key={KEY}&q={CITY}",)
+    data = response.json()
+    print(f"Performing request to Weather API for city {CITY}...")
+    print(f"{data['location']['name']}/{data['location']['country']}"
+          f" {data['location']['localtime']} "
+          f"Weather: {data['current']['temp_c']} "
+          f"Celsius, {data['current']['condition']['text']}")
```

Arsen-hrynevych (changes requested):

BornToLivee:
image

image

Docker Hub screenshot with image

Arsen-hrynevych (approved):
Thanks

PR #542, opened 2024-08-04
Solution

dsokolovskiy:


app/main.py
```diff
@@ -1,6 +1,36 @@
...
+from dotenv import load_dotenv
+
+
+load_dotenv()
+
+API_KEY = os.getenv("API_KEY")
+params = {
+    "key": API_KEY,
+    "q": "Paris",
+    "agi": "no"
```
Arsen-hrynevych, line 15: I don't get it

app/main.py
```diff
@@ -1,6 +1,36 @@
...
+        temp = current_data["temp_c"]
+        text = current_data["condition"]["text"]
+        print(f"{name}/{country} ")
+        print(f"{last_updated} Weather: {temp} Celsius, ")
+        print(f"{text}")
+
+    else:
+        print("Failed to retrieve data.")
+        print(f"Status code: {res.status_code}")
+        print(f"Reason: {res.reason}")
```

Arsen-hrynevych (changes requested):

dsokolovskiy:
image

image

Arsen-hrynevych (changes requested):
I still don't see any link and couldn't find it on Docker Hub

dsokolovskiy:
> I still don't see any link and couldn't find it on Docker Hub

https://hub.docker.com/repositories/hebet

app/main.py
```diff
@@ -1,6 +1,51 @@
...
+        print(f"Error: {e}")
+
+    except ValueError:
+        print("Failed to parse JSON response.")
+
+    except KeyError as e:
+        print(f"Missing key in response data: {e}")
+
+    except Exception as e:
+        print(f"An unexpected error occurred: {e}")
```
Arsen-hrynevych, line 48: Having a single large try block is also not ideal.

Arsen-hrynevych (approved):
Thanks

PR #543, opened 2024-08-04
Solution

SrMusienko:
result


app/main.py
```diff
@@ -1,6 +1,35 @@
...
+        weather = data["current"]["condition"]["text"]
+
+        print(
+            f"{location_name}/{country} "
+            f"{local_time} Weather: {temp_c} Celsius, {weather}"
+        )
+    else:
+        print(
+            f"status code {res.status_code} something went wrong {res.reason}"
+        )
```

Arsen-hrynevych (changes requested):
I see the code, but I couldn’t find the Docker image pushed to Docker Hub. Please provide a link to it. :)

SrMusienko:
https://hub.docker.com/repository/docker/usermusser/weather-api/general

app/main.py
```diff
@@ -1,6 +1,42 @@
...
+            )
+        except ValueError:
+            print("Failed to parse JSON response.")
+
+    except requests.RequestException as e:
+        print(f"Request failed: {e}")
+    except KeyError as e:
+        print(f"Missing key in response data: {e}")
+    except Exception as e:
+        print(f"An unexpected error occurred: {e}")
```
Arsen-hrynevych: Using nested try blocks isn't a good approach, and having a single large try block is also not ideal.

Arsen-hrynevych (changes requested):
Thanks

Arsen-hrynevych (approved):
ChatGPT ;)

PR #546, opened 2024-08-05
Solution v.1

SemenetsSergii:
image


app/main.py
```diff
@@ -1,6 +1,27 @@
...
+    city = data["location"]["name"]
+    country = data["location"]["country"]
+    local_time = data["location"]["localtime"]
+    temp_celsius = data["current"]["temp_c"]
+    weather = data["current"]["condition"]["text"]
+
+    print(
+        f"{city}/{country} "
+        f"{local_time} Weather: {temp_celsius} Celsius, {weather}"
+    )
```
Arsen-hrynevych, line 24: Why Use a try Block Here?

Arsen-hrynevych (changes requested):

app/main.py
SemenetsSergii: docker push sergiisemenets/weather-api

SemenetsSergii:
https://hub.docker.com/repository/docker/sergiisemenets/weather-api/general

docker push sergiisemenets/weather-api:tagname

app/main.py
SemenetsSergii: somesing like that?

def get_weather(key: str, city: str = "Paris") -> None:
    url = "https://api.weatherapi.com/v1/current.json"
    
    try:
        response = requests.get(url, params={"key": key, "q": city})
        response.raise_for_status()  # Raises HTTPError for bad responses (4xx and 5xx)
        
        try:
            weather_data = response.json()
        except ValueError as e:
            print(f"Error parsing JSON response: {e}")
            return

        try:
            location_localtime = weather_data["location"]["localtime"]
            location_name = weather_data["location"]["name"]
            location_country = weather_data["location"]["country"]
            location_weather = weather_data["current"]["condition"]["text"]
            location_wind = weather_data["current"]["wind_mph"]
            temp_c = weather_data["current"]["temp_c"]
            
            print(
                f"[{location_localtime}] {location_name} / {location_country}\n"
                f"Weather: {temp_c} C, {location_weather}, {location_wind} mph\n"
            )
        except KeyError as e:
            print(f"Missing key in JSON response: {e}")
        
    except requests.exceptions.RequestException as e:
        print(f"Network error occurred: {e}")
    except Exception as e:
        print(f"An unexpected error occurred: {e}")

Arsen-hrynevych: Yep

Arsen-hrynevych (approved):
Thanks.

PR #547, opened 2024-08-05
Solution

lilarin:
image
image

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,7 +1,37 @@
-def get_weather() -> None:
-    # write your code here
-    pass
+import os
+import requests
+from dotenv import load_dotenv
+
+
+def get_weather(key: str, city: str = "Paris") -> None:
+    url = "https://api.weatherapi.com/v1/current.json"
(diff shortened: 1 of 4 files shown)


app/main.py
```diff
@@ -1,7 +1,37 @@
...
+        location_weather = weather_data["current"]["condition"]["text"]
+        location_wind = weather_data["current"]["wind_mph"]
+        temp_c = weather_data["current"]["temp_c"]
+
+        print(
+            f"[{location_localtime}] {location_name} / {location_country}\n"
+            f"Weather: {temp_c} C, {location_weather}, {location_wind} mph\n"
+        )
+    else:
+        print(f"Failed to get weather data: {response.status_code}")
```
Arsen-hrynevych, line 28: Why Use a try Block Here?

Arsen-hrynevych (approved):
Thanks

app/main.py
lilarin: Got it

PR #551, opened 2024-08-05
sol

Unlie9:
COMMAND=docker pull unlie9/weather-service

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -16,12 +16,9 @@ def get_weather() -> None:
         weather = response.json()
 
         print(
-            f"Paris/France",
-            f"{weather["current"]["last_updated"]} "
-            f"Weather: "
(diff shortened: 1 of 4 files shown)


app/main.py
```diff
@@ -1,6 +1,25 @@
...
+    response = requests.get(url)
+
+    if response.status_code == 200:
+        weather = response.json()
+
+        print(
+            f"Paris/France {weather['current']['last_updated']} "
+            f"Weather: {weather['current']['temp_c']} "
+            f"Celsius, {weather['current']['condition']['text']}"
+        )
```
Arsen-hrynevych, line 22: Why Use a try Block Here?

Arsen-hrynevych (approved):
Thanks

PR #552, opened 2024-08-05
Solution

b3v3kt0r:


app/main.py
```diff
@@ -1,6 +1,28 @@
...
+    data = json.loads(res.text)
+    location_name = data["location"]["name"]
+    location_country = data["location"]["country"]
+    localtime = data["location"]["localtime"]
+    temp_c = data["current"]["temp_c"]
+    condition_text = data["current"]["condition"]["text"]
+
+    print(f"{location_name}/{location_country}")
+    print(f"{localtime}")
+    print(f"Weather: {temp_c} Celsius, {condition_text}")
```

Arsen-hrynevych (changes requested):

b3v3kt0r:
image

Arsen-hrynevych (approved):
Thanks

PR #556, opened 2024-08-05
Solution

SuskyiVolodymyr:


app/main.py
```diff
@@ -1,6 +1,31 @@
...
+        temp = data["current"]["temp_c"]
+        weather = data["current"]["condition"]["text"]
+        print(f"{city}/{country} {current_time} "
+              f"Weather: {temp} Celsius, {weather}")
+    else:
+        error_message = (
+            response.json().get("error", {})
+            .get("message", "Unknown error occurred")
+        )
+        print("Error:", error_message)
```

Arsen-hrynevych (changes requested):

SuskyiVolodymyr:
link for image: https://hub.docker.com/layers/suskyivolodymyr/py-weather-api/latest/images/sha256:ad7a4a421e50577563280176a51cb06fd229c0815b0f77ce64e7828ac7038dce?uuid=3939ACA1-0691-4280-9E1D-0BC438B450D8

SuskyiVolodymyr:
link for image: https://hub.docker.com/layers/suskyivolodymyr/py-weather-api/latest/images/sha256:ad7a4a421e50577563280176a51cb06fd229c0815b0f77ce64e7828ac7038dce?uuid=3939ACA1-0691-4280-9E1D-0BC438B450D8

Arsen-hrynevych (approved):
Thanks

PR #557, opened 2024-08-05
Solution

YevaYe:


YevaYe:
taxi_task2

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+
+    if response.status_code == 200:
+        weather_data = response.json()
+        current = weather_data["current"]
+        location = weather_data["location"]
+        print(f"Weather for city {location["name"]}, {location["country"]}: ")
+        print(f"Temperature: {current['temp_c']} Celsius")
+        print(f"Condition: {current['condition']['text']}")
+    else:
+        print("Failed to get weather data")
```
Arsen-hrynevych, line 24: Why Use a try Block Here?

Arsen-hrynevych (approved):
Thanks

PR #562, opened 2024-08-10
"Solution"

MaglorFeanorson:
Screenshot_1
Screenshot_4

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM python:3.11-slim
+LABEL maintainer="daniilrog17@gmail.com"
+
+ENV PYTHOUNNBUFFERED=1
+
+WORKDIR /app
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+CMD ["python", "app/main.py"]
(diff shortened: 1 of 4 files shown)


app/main.py
```diff
@@ -1,6 +1,35 @@
...
+        print(
+            f"{city}/{country} "
+            f"{local_time} Weather: {temp_celsius} Celsius, {weather}"
+        )
+    except requests.exceptions.RequestException as e:
+        print(f"An error occurred: {e}")
+    except KeyError as e:
+        print(f"Missing key in the response data: {e}")
+    except Exception as e:
+        print(f"An unexpected error occurred: {e}")
```
Arsen-hrynevych: try block is quite large, which can make it harder to isolate the exact cause of an exception.

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+ENV PYTHOUNNBUFFERED=1
+
+WORKDIR /app
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+CMD ["python", "app/main.py"]
```
Arsen-hrynevych, line 12: Next time, try using docker init.

Arsen-hrynevych (changes requested):
Please provide a link to the pushed Docker image or a screenshot showing that it has been pushed

Dockerfile
MaglorFeanorson: ДЗ

Arsen-hrynevych (approved):
Thanks

PR #565, opened 2024-08-16
Solution

Grotski:


app/main.py
```diff
@@ -1,6 +1,27 @@
...
+
+api_key = os.environ.get("API_KEY")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    city = "Paris"
+    url = f"https://api.weatherapi.com/v1/current.json?key={api_key}&q={city}"
+    response = requests.get(url)
```
vsmutok, line 15: You must check if the request was successful.

vsmutok (changes requested):
you forgot to push the Dockerfile

app/main.py
vasylhnatiuk: do it

vasylhnatiuk: check if status is 200

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+    try:
+        response.status_code == 200
```
vsmutok:
```suggestion
        if response.status_code != 200:
            print("Wrong response status_code")
```

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #570, opened 2024-08-24
Solution

TarasFirst:


app/main.py
```diff
@@ -1,6 +1,28 @@
...
+
+FILTER = "Paris"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+
+    title_request = "Performing request to Weather API for city Paris..."
+    result = requests.get(URL + FILTER)
```
pavlopro: Pass api key and city as second parameter of `get` method

app/main.py
```diff
@@ -1,6 +1,28 @@
...
+    data_paris = result.json()
```
pavlopro: It would be good to check response code to unsure everything is ok, before doing `json()`

pavlopro (changes requested):
Please provide a link to the pushed Docker image or a screenshot showing that it has been pushed

TarasFirst:
https://hub.docker.com/r/gtaras/docker-weather-paris-api/tags
link to image

app/main.py
```diff
@@ -1,6 +1,39 @@
...
+
+    if result.status_code == 200:
+        data_paris = result.json()
+        location = (f"{data_paris['location']['name']}/"
+                    f"{data_paris['location']['country']}")
+        local_time = f"{data_paris['location']['localtime']}"
+        weather = (f" Weather: {data_paris['current']['temp_c']} "
+                   f"Celsius, {data_paris['current']['condition']['text']}")
+        print(title_request)
+        print(location + local_time + weather)
```
vsmutok: combine into one `print()`

app/main.py
```diff
@@ -1,6 +1,39 @@
...
+    else:
+        print(f"Failed to retrieve data: {result.status_code}")
+        print(f"Response: {result.text}")
```
vsmutok: combine into one `print()`

.dockerignore
```diff
@@ -0,0 +1 @@
+venv
```
vsmutok, line 1: What about other files that should not be included?

vsmutok (changes requested):

.dockerignore
TarasFirst: > What about other files that should not be included?

Ok. Which files?

app/main.py
TarasFirst: > combine into one `print()`

I think, this code more readable and understandable, but if you think, that, this way will be better, ok.
But i dont understand, why

.dockerignore
vsmutok: it looks better now

vsmutok (approved):
Great!

PR #575, opened 2024-08-30
Solution

MykhailoIvchenko:


Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+LABEL maintainer="ivchenkomykhailo@gmail.com"
+
+ENV PYTHONUNBUFFERED 1
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py", "0.0.0.0:8000"]
```
vsmutok: Why you need `0.0.0.0:8000`?

vsmutok (changes requested):

Dockerfile
MykhailoIvchenko: Because I haven't understood that I don't need it if I do not run a server))
Fixed

vsmutok (approved):
Great!

PR #584, opened 2024-09-27
solution

Fedot0v:


app/main.py
```diff
@@ -1,6 +1,38 @@
...
+            "Please set the API_KEY environment variable."
+        )
+
+    base_url = "https://api.weatherapi.com/v1"
+    method = "current.json"
+    location = "Paris"
+
+    url = f"{base_url}/{method}?q={location}&key={api_key}"
+
+    response = requests.get(url)
```
pavlopro: use `params` as second argument of `get` method to pass parameters

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #594, opened 2024-10-07
Solution

dartomOOv:


.dockeringore
```diff
@@ -0,0 +1,3 @@
+.venv
+result.png
+Dockerfile
```
DSahalatyi, line 3: I think you may add some more items here

app/main.py
```diff
@@ -1,6 +1,29 @@
+import os
+
+import requests
+
+
+URL = "https://api.weatherapi.com/v1/current.json?"
```
DSahalatyi, line 6: URL can be split into base_url and endpoint.

app/main.py
```diff
@@ -1,6 +1,29 @@
...
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    print(f"Performing request to Weather API for city {CITY}...")
+
+    res = requests.get(
+        f"{URL}key={API_KEY}&q={CITY}"
+    )
```
DSahalatyi, line 16: You can pass parameters using params dict as an argument

DSahalatyi (approved):

.dockeringore
Arsen-hrynevych: You definitely need to add more files.

Arsen-hrynevych (approved):
ok

PR #596, opened 2024-10-07
Solution

AndriiHamasa:


.dockerignore
```diff
@@ -0,0 +1 @@
+.venv
```
OleksiukStepan: You have many more file in your main dir that you don't need to push into Docker

OleksiukStepan (approved):

.dockerignore
AndriiHamasa: Yeah, I've created and forget about this file

Arsen-hrynevych (approved):
Ok

PR #601, opened 2024-10-08
Solution

pashawarganov:
image


.dockerignore
```diff
@@ -0,0 +1 @@
+.venv
```
ValeriiaKyr, line 1: I think you need to add more files

ValeriiaKyr (approved):

app/main.py
```diff
@@ -1,6 +1,39 @@
+import os
+from datetime import datetime
+
+import requests
+from dotenv import load_dotenv
+
+load_dotenv()
+
+API_KEY = os.environ.get("API_KEY")
+API_URL = f"https://api.weatherapi.com/v1/current.json?key={API_KEY}&q="
```
oliasymak24: If you are using fstrings then this `API_URL` is not a constant any more and will change depending on API_KEY

oliasymak24 (changes requested):
You also must modify READme file)

oliasymak24 (approved):

PR #606, opened 2024-10-08
Task Solution

achalaidiuk:


app/main.py
```diff
@@ -1,6 +1,36 @@
...
+BASE_URL = "https://api.weatherapi.com/v1/"
+CITY = "Paris"
+
+
+def get_weather() -> str:
+    payload = {"key": API_KEY, "q": CITY}
+    get_weather_url = BASE_URL + "current.json"
+    response = requests.get(get_weather_url, params=payload)
+    print(payload)
+    print(get_weather_url)
```
HalynaRohatska: I think you don't need these prints

HalynaRohatska (approved):

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.12.3
```
oliasymak24: Read the checklist please

app/main.py
oliasymak24: Delete these prints

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #608, opened 2024-10-08
Solution

dpiuro:
forecast


.dockerignore
DSahalatyi, line 1: Might add even more items here.

DSahalatyi (approved):

PR #609, opened 2024-10-09
Solution

Poznanskyi:


Maksym-Turenko (changes requested):
You need to specify the command in README.md so that mentor can make a pull of your image from dockerhub.

PR #610, opened 2024-10-09
implement py-docker-weather-api

Maksym-Turenko:


mnezbrytska (approved):
It may make sense to check the status of the response before accessing the data.(in main.py)

PR #615, opened 2024-10-10
Solution

Bogdan-Rylov:


.dockerignore
```diff
@@ -0,0 +1,4 @@
+.venv
+.env
+*.pyc
+__pycache__/
```
Arsen-hrynevych, line 4: Please add more files into dockerignore for better perfomance

Arsen-hrynevych (approved):
Ok

PR #646, opened 2024-12-10
Solution

vladyslav-tmf:
Screenshot_1

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,48 @@
-FROM python:3.12-slim
+# syntax=docker/dockerfile:1
 
+# Comments are provided throughout this file to help you get started.
+# If you need more help, visit the Dockerfile reference guide at
+# https://docs.docker.com/go/dockerfile-reference/
+
+# Want to help us make this template better? Share your feedback here: https://forms.gle/ybq9Krt8jtBL3iCk7
+
+ARG PYTHON_VERSION=3.12
+FROM python:${PYTHON_VERSION}-slim AS base
 LABEL maintainer="vladislav.tmf@gmail.com"
 
+# Prevents Python from writing pyc files.
+ENV PYTHONDONTWRITEBYTECODE=1
+
+# Keeps Python from buffering stdout and stderr to avoid situations where
+# the application crashes without emitting any logs due to buffering.
 ENV PYTHONUNBUFFERED=1
 
-WORKDIR /src
+WORKDIR /app
 
+# Copy requirements first to leverage Docker cache
 COPY requirements.txt requirements.txt
-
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+WORKDIR /src
+
+COPY requirements.txt requirements.txt
+
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
```
# syntax=docker/dockerfile:1

ARG PYTHON_VERSION=3.12

ENV PYTHONDONTWRITEBYTECODE=1

ENV PYTHONUNBUFFERED=1

WORKDIR /app

ARG UID=10001
RUN adduser \
    --disabled-password \
    --gecos "" \
    --home "/app" \
    --shell "/sbin/nologin" \
    --no-create-home \
    --uid "${UID}" \
    appuser

USER appuser

COPY ./app /app/

EXPOSE 8000

# Run the application.
CMD ["python", "app/main.py"]
```

app/main.py
```diff
@@ -1,6 +1,38 @@
+import os
+
+import requests
+
```
Arsen-hrynevych, line 4: 2 blank lines after imports

app/main.py
```diff
@@ -1,6 +1,38 @@
...
+API_URL = "http://api.weatherapi.com/v1/current.json"
```
Arsen-hrynevych: You need to generate API_KEY for using Weather API, but it shouldn't be hard-coded in main.py.

app/main.py
```diff
@@ -1,6 +1,38 @@
...
+CITY = "Paris"
```
Arsen-hrynevych: it shouldn't be hard-coded in main.py.

app/main.py
```diff
@@ -1,6 +1,38 @@
...
+    current = data["current"]
+    local_time = location["localtime"]
+    current_temp = current["temp_c"]
+    current_conditions = current["condition"]["text"]
+
+    print(f"Performing request to Weather API for city {CITY}...")
+    print(
+        f"{CITY}/{country} {local_time} Weather: {current_temp} Celsius, "
+        f"{current_conditions}"
+    )
```
Arsen-hrynevych: This doesn't follow the Single Responsibility Principle. You could split it into two functions and use a try-except block for each. Also, consider changing the naming of the response and other variables. Thanks!

Arsen-hrynevych (changes requested):
Please address the issues, and if you have any questions, write them next to the comments.

Arsen-hrynevych (approved):
Good

PR #652, opened 2024-12-12
Solution

VladyslavBon:


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.11.6-alpine3.18
+
+ENV PYTHONUNBUFFERED 1
+
+WORKDIR app/
+
+COPY . .
+RUN pip install -r requirements.txt
+
+CMD ["python", "app/main.py"]
```
Arsen-hrynevych: The Dockerfile does not follow best practices
https://docs.docker.com/build/building/best-practices/

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+    country = result["location"]["country"]
+    local_time = result["location"]["localtime"]
+    temperature = result["current"]["temp_c"]
+    condition = result["current"]["condition"]["text"]
+
+    print("Performing request to Weather API for city Paris...")
+    print(
+        f"{city}/{country} {local_time} "
+        f"Weather: {temperature} Celsius, {condition}"
+    )
```
Arsen-hrynevych: Please add try exception block, also you could split into functions

app/main.py
```diff
@@ -1,6 +1,27 @@
+import os
+
+import requests
+
+
+API_KEY = os.getenv("api_key")
+URL = "https://api.weatherapi.com/v1/current.json?"
+FILTERING = "Paris"
```
Arsen-hrynevych: If you set api_key as an environment variable, why not do the same for URL and FILTERING?

Arsen-hrynevych (changes requested):
Mostly correct/

app/main.py
VladyslavBon: It's our task condition:
  You need to generate API_KEY for using Weather API, but it shouldn't be hard-coded in main.py.
  You should use environment variables for that purpose. So, read API_KEY from environment. And to pass environment variables to docker container use -e flag.

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+WORKDIR /app
+COPY requirements.txt .
+RUN pip install --no-cache-dir -r requirements.txt
+
+RUN adduser -D myuser
+USER myuser
+
+COPY . .
+
+CMD ["python3", "app/main.py"]
```
```
# syntax=docker/dockerfile:1

ARG PYTHON_VERSION=3.12

ENV PYTHONDONTWRITEBYTECODE=1

ENV PYTHONUNBUFFERED=1

WORKDIR /app

ARG UID=10001
RUN adduser \
    --disabled-password \
    --gecos "" \
    --home "/app" \
    --shell "/sbin/nologin" \
    --no-create-home \
    --uid "${UID}" \
    appuser

USER appuser

COPY ./app /app/

EXPOSE 8000

# Використовуємо змінну середовища для вибору запуску

# Run the application.
CMD python app/main.py
```

app/main.py
Arsen-hrynevych: First, in the naming convention, you should use os.getenv("API_KEY"). Second, if you pass API_KEY as an environment variable, ensure it is correctly set. Regarding other fields, I’m confused about why I can’t decide which city I want to see. I want to have a single configuration where I can input the desired city. The same applies to the URL—what happens if it changes? What should we do then?

app/main.py
```diff
@@ -1,6 +1,33 @@
...
+    except KeyError as e:
+        return f"Missing data in response: {e}"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    result = requests.get(f"{URL}key={API_KEY}&q={FILTERING}").json()
+
+    print(parse_weather_data(result))
```
Arsen-hrynevych: Why try exception for parse_weather_data but not for get_weather?

app/main.py
```diff
@@ -1,6 +1,33 @@
...
+            f"Weather: {temperature} Celsius, {condition}"
+        )
+    except KeyError as e:
+        return f"Missing data in response: {e}"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    result = requests.get(f"{URL}key={API_KEY}&q={FILTERING}").json()
```
Arsen-hrynevych: using .json in the end is risky. Better approach to split it into 2 variables.

Arsen-hrynevych: Result name: Does this describe the result of the request? Partially yes, but it does not describe the data inside it. A better name would be weather_response.

Arsen-hrynevych (changes requested):
Please address the issues, and if you have any questions, write them next to the comments.

app/main.py
```diff
@@ -1,6 +1,37 @@
+import os
+
+import requests
+
+
+API_KEY = os.getenv("API_KEY")
+URL = "https://api.weatherapi.com/v1/current.json?"
+FILTERING = "Paris"
```
Arsen-hrynevych, line 8: make it dymanic not static.

app/main.py
```diff
@@ -1,6 +1,37 @@
...
+        country = weather_data["location"]["country"]
+        local_time = weather_data["location"]["localtime"]
+        temperature = weather_data["current"]["temp_c"]
+        condition = weather_data["current"]["condition"]["text"]
+
+        return (
+            f"{city}/{country} {local_time} "
+            f"Weather: {temperature} Celsius, {condition}"
+        )
+    except KeyError as e:
```
Arsen-hrynevych, line 23: use more descriptive naming e -> exc or e -> exception:
I prefer exc

.dockerignore
```diff
@@ -0,0 +1,2 @@
+venv
+Dockerfile
```
Arsen-hrynevych, line 2: Please add .dockerignore also

Arsen-hrynevych (approved):
ok

PR #659, opened 2024-12-17
Develop

Bondzik-S:


app/main.py
```diff
@@ -1,7 +1,34 @@
-def get_weather() -> None:
-    # write your code here
-    pass
+import os
+import requests
+
```
alauda6, line 3: I think you missed to import load_dotenv to properly import API_KEY from your .env file

alauda6 (approved):

PR #660, opened 2024-12-18
solution

alauda6:
Screenshot 2024-12-18 at 13 27 00

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM python:3.13.0-slim
+LABEL authors="serhiik"
+
+ENV PYTHONUNNBUFFERED=1
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+EXPOSE 8000
+
+CMD ["python", "app/main.py"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.13.0-slim
+FROM python:3.11.0-slim
 LABEL authors="serhiik"
 
 ENV PYTHONUNNBUFFERED=1
--- a/app/main.py
+++ b/app/main.py
@@ -1,6 +1,37 @@
-def get_weather() -> None:
-    # write your code here
-    pass
+import os
+
+import requests
+
+from dotenv import load_dotenv
+
+load_dotenv()
+
+BASE_URL = "http://api.weatherapi.com/v1/current.json"
+
+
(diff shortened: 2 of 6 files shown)


Atikiho (approved):

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ENV PYTHONUNNBUFFERED=1
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+EXPOSE 8000
+
+CMD ["python", "app/main.py"]
```
Arsen-hrynevych, line 13:
```
# syntax=docker/dockerfile:1

ARG PYTHON_VERSION=3.12

ENV PYTHONDONTWRITEBYTECODE=1

ENV PYTHONUNBUFFERED=1

WORKDIR /app

ARG UID=10001
RUN adduser \
    --disabled-password \
    --gecos "" \
    --home "/app" \
    --shell "/sbin/nologin" \
    --no-create-home \
    --uid "${UID}" \
    appuser

USER appuser

COPY main.py test_main.py /app/

EXPOSE 8000

# Run the application.
CMD python app/main.py
```

app/main.py
```diff
@@ -1,6 +1,37 @@
...
+
+        city_name = data["location"]["name"]
+        country = data["location"]["country"]
+        localtime = data["location"]["localtime"]
+        temperature = data["current"]["temp_c"]
+        condition = data["current"]["condition"]["text"]
+        print(f"{city_name}/{country} {localtime} "
+              f"Weather: {temperature}°C, {condition}")
+    else:
+        print(f"Error {response.status_code}: {response.text}")
```
Arsen-hrynevych, line 34: Please split this into functions.

app/main.py
```diff
@@ -1,6 +1,37 @@
...
+
+from dotenv import load_dotenv
+
+load_dotenv()
+
+BASE_URL = "http://api.weatherapi.com/v1/current.json"
+
+
+def get_weather(city: str = "Paris") -> None:
+    api_key = os.getenv("API_KEY")
```
Arsen-hrynevych, line 13: api_key should be renamed to API_KEY because it represents a constant. Additionally, it should be declared beneath BASE_URL for better organization.

app/main.py
Arsen-hrynevych, line 34: Use try exception block

Arsen-hrynevych (changes requested):
Mostly correct

PR #674, opened 2024-12-19
Solution

Den-k0:
image
image


sumseer (approved):

.dockerignore
```diff
@@ -0,0 +1,2 @@
+.venv
+.env
```
Arsen-hrynevych, line 2: Please add more files

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:3.12-slim
+LABEL maintainer="denys.kotovskiy@gmail.com"
+
+ENV PYTHONUNBUFFERED 1
+
+WORKDIR /app
```
Arsen-hrynevych, line 6: Use docker init to create a Dockerfile. Right now, your file isn't following best practices.

```
# syntax=docker/dockerfile:1

ARG PYTHON_VERSION=3.12

ENV PYTHONDONTWRITEBYTECODE=1

ENV PYTHONUNBUFFERED=1

WORKDIR /app

ARG UID=10001
RUN adduser \
    --disabled-password \
    --gecos "" \
    --home "/app" \
    --shell "/sbin/nologin" \
    --no-create-home \
    --uid "${UID}" \
    appuser

USER appuser

COPY main.py test_main.py /app/

EXPOSE 8000

# Run the application.
CMD python app/main.py
```

app/main.py
```diff
@@ -1,6 +1,27 @@
+import os
+import requests
+from dotenv import load_dotenv
```
Arsen-hrynevych: Imports should be grouped and separated by a blank line according to standard library, external libraries, and custom imports. This practice enhances both code readability and maintainability. 
Read more about topic [here](https://peps.python.org/pep-0008/#imports)

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+
+load_dotenv()
+
+API_KEY = os.getenv("API_KEY")
```
Arsen-hrynevych: Please change the method from optional to strict, as it should raise a KeyError if the key is not provided

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+URL = "http://api.weatherapi.com/v1/current.json"
+FILTERING = "Paris"
```
Arsen-hrynevych, line 10: Make it dynamic

app/main.py
```diff
@@ -1,6 +1,27 @@
+import os
+import requests
+from dotenv import load_dotenv
+
+load_dotenv()
+
+API_KEY = os.getenv("API_KEY")
+URL = "http://api.weatherapi.com/v1/current.json"
```
Arsen-hrynevych, line 9: The same apply here

app/main.py
```diff
@@ -1,6 +1,27 @@
...
+    response.raise_for_status()
+    data = response.json()
+
+    print(
+        f"{data['location']['name']}/"
+        f"{data['location']['country']} "
+        f"{data['location']['localtime']} "
+        f"Weather: {data['current']['temp_c']} Celsius, "
+        f"{data['current']['condition']['text']}"
+    )
```
Arsen-hrynevych: Split into functions, use try except block.

Arsen-hrynevych (changes requested):
Mostly correct.

Arsen-hrynevych (changes requested):
Has something changed?

PR #680, opened 2024-12-23
Solution

zaietsmo:


.dockerignore
```diff
@@ -0,0 +1,6 @@
+__pycache__/
+*.pyc
+.env
+.idea
+*.log
+*.tmp
```
Arsen-hrynevych: Please include more files

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM python:3.9-slim
+WORKDIR /app
+COPY requirements.txt ./
+RUN pip install --no-cache-dir -r requirements.txt
+COPY app/ ./app
+ENV API_KEY=""
+CMD ["python", "app/main.py"]
```
```
# syntax=docker/dockerfile:1

ARG PYTHON_VERSION=3.12

ENV PYTHONDONTWRITEBYTECODE=1

ENV PYTHONUNBUFFERED=1

WORKDIR /app

ARG UID=10001
RUN adduser \
    --disabled-password \
    --gecos "" \
    --home "/nonexistent" \
    --shell "/sbin/nologin" \
    --no-create-home \
    --uid "${UID}" \
    appuser

# Download dependencies as a separate step to take advantage of Docker's caching.
# Leverage a cache mount to /root/.cache/pip to speed up subsequent builds.
# Leverage a bind mount to requirements.txt to avoid having to copy them into
# into this layer.
RUN --mount=type=cache,target=/root/.cache/pip \
    --mount=type=bind,source=requirements.txt,target=requirements.txt \
    python -m pip install -r requirements.txt

USER appuser

COPY . .

# Run the application.
CMD python app/main.py

```

app/main.py
```diff
@@ -1,7 +1,27 @@
...
+        return response.json()
+    else:
+        return None
+
+
+# Main function
 if __name__ == "__main__":
-    get_weather()
+    city = "San Francisco"
+    weather_data = get_weather(city)
```
Arsen-hrynevych: Make it dynamic and move the city parameter next to the URL and API key.

app/main.py
```diff
@@ -1,7 +1,27 @@
...
+url = "http://api.weatherapi.com/v1/current.json"
 
+
+def get_weather(city: str) -> dict | None:
+    params = {"key": API_KEY, "q": city}
+    response = requests.get(url, params=params)
+    if response.status_code == 200:
+        return response.json()
+    else:
+        return None
```
Arsen-hrynevych: Use try exception block for handle exception

app/main.py
```diff
@@ -1,7 +1,27 @@
-def get_weather() -> None:
-    # write your code here
-    pass
+import os
+import requests
 
+API_KEY = os.getenv("API_KEY")
```
Arsen-hrynevych, line 4: API_KEY is optional? Or shoud we raise KeyError if API_KEY is not provided

Arsen-hrynevych (changes requested):
Mostly correct

.dockerignore
```diff
@@ -0,0 +1,411 @@
...
+*.cb7
+*.cbt
+*.cba
+*.cbr
+*.cbz
+*.cb7
+*.cbt
+*.cba
+*.cbr
+*.cbz
```
Arsen-hrynevych, line 129: What happened? Why are there so many files here?

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+# Create a non-privileged user
+ARG UID=10001
+RUN adduser \
+    --disabled-password \
+    --gecos "" \
+    --home "/nonexistent" \
+    --shell "/sbin/nologin" \
+    --no-create-home \
+    --uid "${UID}" \
+    appuser
```
Arsen-hrynevych, line 26: There is a logic issue with the folders.

app/main.py
```diff
@@ -1,7 +1,31 @@
...
+if not API_KEY:
+    raise KeyError("API_KEY not provided")
```
Arsen-hrynevych, line 6: For this purpose, you could just use API_KEY = os.environ["API_KEY"]

app/main.py
```diff
@@ -1,7 +1,31 @@
...
+url = "http://api.weatherapi.com/v1/current.json"
+CITY = "San Francisco"
```
Arsen-hrynevych, line 8: Why is this being hardcoded?

app/main.py
```diff
@@ -1,7 +1,31 @@
...
+
+def get_weather(city: str) -> dict | None:
+    params = {"key": API_KEY, "q": city}
+    try:
+        response = requests.get(url, params=params)
+        response.raise_for_status()
+        return response.json()
+    except requests.exceptions.RequestException as e:
+        print(f"Error fetching weather data: {e}")
+        return None
```
Arsen-hrynevych, line 19: Using a single general exception is bad practice.

Arsen-hrynevych (changes requested):
Mostly correct

PR #682, opened 2024-12-23
Solution

molodsh1y1:
Screenshot 2024-12-23 085121


.dockerignore
```diff
@@ -0,0 +1,6 @@
+**/.idea/
+**/.vscode/
+**/.DS_Store
+**/venv/
+**/.pytest_cache/
+**/**__pycache__/
```
Arsen-hrynevych: Please include more files

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+ENV PYTHONUNBUFFERED 1
+
+WORKDIR /app
+
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
```
`# syntax=docker/dockerfile:1

ARG PYTHON_VERSION=3.12

ENV PYTHONDONTWRITEBYTECODE=1

ENV PYTHONUNBUFFERED=1

WORKDIR /app

ARG UID=10001
RUN adduser \
    --disabled-password \
    --gecos "" \
    --home "/nonexistent" \
    --shell "/sbin/nologin" \
    --no-create-home \
    --uid "${UID}" \
    appuser

# Download dependencies as a separate step to take advantage of Docker's caching.
# Leverage a cache mount to /root/.cache/pip to speed up subsequent builds.
# Leverage a bind mount to requirements.txt to avoid having to copy them into
# into this layer.
RUN --mount=type=cache,target=/root/.cache/pip \
    --mount=type=bind,source=requirements.txt,target=requirements.txt \
    python -m pip install -r requirements.txt

USER appuser

COPY . .

# Run the application.
CMD python app/main.py
`
```

app/main.py
```diff
@@ -1,6 +1,57 @@
...
+    JSONDecodeError,
+    RequestException
+)
+
+from dotenv import load_dotenv
+
+load_dotenv()
+
+URL = "http://api.weatherapi.com/v1/current.json?"
+FILTERING = "Paris"
```
Arsen-hrynevych, line 15: make it dynamic also

app/main.py
```diff
@@ -1,6 +1,57 @@
+import os
+import requests
+from requests.exceptions import (
+    ConnectionError,
+    JSONDecodeError,
+    RequestException
+)
+
+from dotenv import load_dotenv
```
Arsen-hrynevych: Imports should be grouped and separated by a blank line according to standard library, external libraries, and custom imports. This practice enhances both code readability and maintainability. 
Read more about topic [here](https://peps.python.org/pep-0008/#imports)

app/main.py
```diff
@@ -1,6 +1,57 @@
...
+
+load_dotenv()
+
+URL = "http://api.weatherapi.com/v1/current.json?"
+FILTERING = "Paris"
+API_KEY = os.getenv("API_KEY")
```
Arsen-hrynevych: os.getenv doesn't raise a KeyError exception if the key is not provided. You can use os.environ and remove the if statement.

app/main.py
```diff
@@ -1,6 +1,57 @@
...
+    except ConnectionError as exc:
+        print(f"Connection error: {exc}")
+    except JSONDecodeError as exc:
+        print(f"JSON decoding error: {exc}")
+    except RequestException as exc:
+        print(f"Request error: {exc}")
+    return {}
+
+
+def print_weather_data(data: dict) -> None:
```
Arsen-hrynevych: Please add short docstings

Arsen-hrynevych (changes requested):
Mostly correct. Thanks

AnyoneClown (approved):
LGTM

PR #710, opened 2025-02-16
Solution

taraskhom:


Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM python:3.10-slim
```
4ikkamony, line 1: https://hub.docker.com/_/python
You can use some version of alpine instead of slim, it's ~ 100 mb smaller

4ikkamony (approved):
Nice! Godly exception catching

PR #719, opened 2025-02-18
Docker weather API for Paris

Sichkaridze:
Screenshot 2025-02-18 at 15 23 05


Dogober (approved):
Perfect job. You add error handler to get weather function and use get method for avoiding errors when you work with json.

PR #731, opened 2025-02-19
Implemented solution

kuinorymes:


Nattalli (changes requested):
OK but what about that line in README.md:
```
You must *modify* this line with correct command to pull your image:
  COMMAND=`docker pull <YOUR_DOCKER_ID/YOUR_IMAGE_NAME>`.

Nattalli (changes requested):
so you had to crate docker image and write it's name

kuinorymes:
Знімок екрана 2025-02-21 191455
i hope i got it right:)
Took a screenshot also

> so you had to crate docker image and write it's name

Nattalli (approved):

PR #744, opened 2025-03-16
Develop

uzlss:


uzlss:
@mateacademy-ai-mentor
1. python alpine is super slim.
2. all constant names are written in uppercase.

PR #764, opened 2025-04-24
Anton Zhukov - Solution

windings-lab:


windings-lab:
@mateacademy-ai-mentor 
1. Workdir set to /app. That means when I `COPY /app .`, i am copying from my local directory called app to the WORKDIR directory called app. I only need to copy main.py that's all.
2. I am correctly checking in the code of get_weather if API_KEY is wrong using try, except

PR #780, opened 2025-05-06
solution

yarzhuravlov:


yarzhuravlov:
@mateacademy-ai-mentor I've reviewed your comments and I'd like to clarify a few points that were misunderstood
Issue: The CMD instruction uses 'python app/main.py', but since WORKDIR is already set to 'app', the correct path should be 'main.py'.

This CMD instruction is correct because main.py is in app directory that copied to app WORKDIR. so path to the main.py is app/app/main.py

yarzhuravlov:
@mateacademy-ai-mentor  I've reviewed your comments and I'd like to clarify a few points that were misunderstood
Issue: The CMD instruction uses 'python app/main.py', but since WORKDIR is already set to 'app', the correct path should be 'main.py'.

This CMD instruction is correct because main.py is in app directory that copied to app WORKDIR. so path to the main.py is app/app/main.py

if we change CMD to ["python", "main.py"] we will have such error `python: can't open file '/app/main.py': [Errno 2] No such file or directory`

PR #812, opened 2025-06-01
Solution

Fantom245:


app/main.py
```diff
@@ -1,7 +1,23 @@
+import requests
+
+
 def get_weather() -> None:
```
vsmutok: Issue: The function 'get_weather' is annotated to return None (-> None), but it actually returns a string in both success and error cases. You should update the return type annotation to '-> str' to accurately reflect the function's behavior.

app/main.py
```diff
@@ -1,7 +1,23 @@
...
-    # write your code here
-    pass
+
+    API_KEY = "b47adef0ac6d4d919b5131859253105"
```
vsmutok: You must not store API keys or any confidential data in a public GitHub repository.

app/main.py
```diff
@@ -1,7 +1,23 @@
...
+    url = f"http://api.weatherapi.com/v1/current.json?key={API_KEY}&q={CITY}&lang=en"
+
+    response = requests.get(url)
+    if response.status_code == 200:
+        data = response.json()
+        location = data["location"]["name"]
+        temp = data["current"]["temp_c"]
+        condition = data["current"]["condition"]["text"]
+        return f"Weather in {location}: {condition}, {temp}°C"
+    else:
```
vsmutok: `else` redundant here

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #877, opened 2025-09-10
Solution

yagginov:
image


Artcrafterrra (approved):
LGTM

yagginov:
@mateacademy-ai-mentor 

1. We don't need a Dockerfile in the docker image itself. That's why it was added to dockerignore.
2. alpine python builds are thinner than slim.

yagginov:
@mateacademy-ai-mentor 

These remarks don’t make sense:
- .dockerignore: Having Dockerfile listed there doesn’t break the build in my case. The image builds successfully, and removing it from the build context is not an issue here.
- Dockerfile base image: Alpine is already thinner than slim variants, so switching to slim would not make the image “lighter.”
- requirements.txt: The file is present in the repo, so the claim that it’s missing is incorrect.

PR #878, opened 2025-09-11
Solution

Artcrafterrra:
Screenshot 2025-09-11 125916


Otst0y (approved):

Artcrafterrra:
@mateacademy-ai-mentor 1. It's not necessary to add a .dockerignore file to the .dockerignore file because its size is just a few kilobytes. But I`ll do that.
2. requirements.txt already exists in my repo, and its main purpose is to install dependencies.
3. The main task is to get the weather in Paris. And in code we already haw a varible CITY, so it`s not nessesary to do that.

PR #890, opened 2025-09-11
Solution

viannik:


app/main.py
```diff
@@ -1,6 +1,35 @@
+import os
+import requests
+import sys
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    api_key = os.getenv("API_KEY")
```
SensibleN00B: image

it is a good practice to bring this before finction like in checklist

SensibleN00B (changes requested):
please change url, api_key and add filtering constants like in checklist

viannik:
@mateacademy-ai-mentor 
Thanks for the feedback. I've reviewed the points and here are my clarifications:

1. Regarding the base image: I chose python:3.13-alpine because it is significantly smaller than the slim variants, which aligns with the requirement to keep the image as thin as possible.
2. Regarding requirements.txt: The requirements.txt file, which includes the requests library, was created and is present in the project directory. The build runs successfully without any errors.
3. Regarding Docker Hub: The image has been published to Docker Hub, and the README file has been updated with the correct command: docker pull sadronu/weather-api.

I believe all the requirements from the checklist have been met. Let me know if there's anything else.

AnyoneClown (approved):
Next time divide logic into different functions such as fetch_weather, parse_weather, print_weather

PR #895, opened 2025-09-11
Solution

DenisPrudius:


app/main.py
```diff
@@ -1,6 +1,33 @@
...
+    }
+    response = requests.get(url, params=params)
+    response.raise_for_status()
+    data = response.json()
+
+    city = data["location"]["name"]
+    temp_c = data["current"]["temp_c"]
+    condition = data["current"]["condition"]["text"]
+
+    print(f"Weather in {city}: {temp_c}°C, {condition}")
```
AnyoneClown, line 36: Divide logic into different functions such as `fetch_weather`, `parse_weather`, `print_weather`

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+ENV PYTHONUNBUFFERED=1
+
+WORKDIR /app
+COPY requirements.txt requirements.txt
+RUN pip install -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
AnyoneClown, line 12: Use best practice for Dockerfile. Use docker init - https://docs.docker.com/reference/cli/docker/init/

AnyoneClown (changes requested):

app/main.py
DenisPrudius: Done ✅

PR #896, opened 2025-09-12
Solution

Irahromska:


AnyoneClown (approved):
Read about `docker init`  - https://docs.docker.com/reference/cli/docker/init/

PR #902, opened 2025-09-16
Solution

Arch0998:
Screenshot 2025-09-26 184508

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,51 @@
-FROM python:3.12-slim
-LABEL maintainer="arthur.oleinikov.py@gmail.com"
+# syntax=docker/dockerfile:1
 
+# Comments are provided throughout this file to help you get started.
+# If you need more help, visit the Dockerfile reference guide at
+# https://docs.docker.com/go/dockerfile-reference/
+
+# Want to help us make this template better? Share your feedback here: https://forms.gle/ybq9Krt8jtBL3iCk7
+
+ARG PYTHON_VERSION=3.12.0
+FROM python:${PYTHON_VERSION}-slim as base
+
+# Prevents Python from writing pyc files.
+ENV PYTHONDONTWRITEBYTECODE=1
+
+# Keeps Python from buffering stdout and stderr to avoid situations where
+# the application crashes without emitting any logs due to buffering.
 ENV PYTHONUNBUFFERED=1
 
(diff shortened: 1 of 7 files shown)


app/main.py
```diff
@@ -1,6 +1,18 @@
...
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    url = "http://api.weatherapi.com/v1/"
+    filtering = "Paris"
+
+    result = requests.get(url + f"current.json?key={API_KEY}&q={filtering}")
+
+    print(result.json())
```
AnyoneClown: Read `checklist.md`

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ENV PYTHONUNBUFFERED=1
+
+WORKDIR /app
+
+COPY requirements.txt ./
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY . .
+
+CMD ["python", "app/main.py"]
```
AnyoneClown: Refactor with `docker init`

AnyoneClown (changes requested):

app/main.py
Arch0998: @AnyoneClown
Yes, my first commit had a "constant", but flake8 rejected it and I had to write it the way I did :D
You can check my first commit that was rejected)

L1nk3rrr: @Arch0998 that because you placed contasant inside function, but it must be our, as you did it with API key.

app/main.py
```diff
@@ -1,6 +1,18 @@
...
+API_KEY = os.environ.get("API_KEY")
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    url = "http://api.weatherapi.com/v1/"
+    filtering = "Paris"
+
+    result = requests.get(url + f"current.json?key={API_KEY}&q={filtering}")
```
L1nk3rrr: also `requests` have posibility to set query params via `params`, not inline

app/main.py
L1nk3rrr: I think it can be parsed and printed in a better way, than simply show to user a json

L1nk3rrr (changes requested):
Minor changes required

AnyoneClown (approved):
LGTM

PR #1109, opened 2026-08-08
Solution docker weather api

Oh-no-dotcom:


app/main.py
```diff
@@ -1,6 +1,17 @@
...
+
+CITY = "Paris"
+URL = "http://api.weatherapi.com/v1/current.json"
+
+
 def get_weather() -> None:
-    # write your code here
-    pass
+    api_key = os.environ.get("API_KEY")
+    response = requests.get(f"{URL}?key={api_key}&q={CITY}")
```
vsmutok, line 10: You should make sure that the request was successful and that `API_KEY` is not `None`.

app/main.py
```diff
@@ -1,6 +1,17 @@
...
+    data = response.json()
+    print(data["location"]["name"])
+    print(data["current"]["temp_c"])
+    print(data["current"]["condition"]["text"])
```
vsmutok, line 14: you can combine this into one `print`

vsmutok (changes requested):

